PR #5, opened 2015-06-21
returning an indiviudal resource should be a map, not a list

mitchellhenke:
if I read correctly
http://jsonapi.org/format/#fetching-resources-responses

@jeregrine


jeregrine:
👍

PR #16, opened 2015-09-02
Ensure PlugResponseContentType always returns the conn

adamesque:
I'd have thought Dialyzer would catch this (because the Plug behvaiour seems to include a spec for the return value of a `register_before_send` callback), but it didn't.

Anyway, pretty important we always get a `conn` back from this. ;)


jeregrine:
Thank you!!!

PR #22, opened 2015-12-17
Ecto upgrade

jakerobers:
The only breaking change from 0.11 to 1.0 is that `Ecto.Associations.loaded?` was changed to `Ecto.assoc_loaded?`


jeregrine:
Great work!

PR #25, opened 2015-12-17
Added a basic travis.yml.

jakerobers:
@jeregrine I authenticated to travisci with my account, but I could not enable CI for your repo. I believe this is something you need to do.

The default command for test execution is `mix test` with travisci, so I think we should be good to go.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,3 @@
+language: elixir
+elixir:
+  - 1.1.1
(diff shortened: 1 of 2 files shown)


jeregrine:
@jakerobers wanna add this to the readme too?

```md
[![Build Status](https://travis-ci.org/jeregrine/jsonapi.svg)](https://travis-ci.org/jeregrine/jsonapi)
```

Issue #26, opened 2015-12-21
Errors in tests in example app, related to includes/relationships


alexjp:
Hi,

When in branch overhaul, commenting the includes and relationships if the views, make all tests pass.
When enabling includes and relationships in views ( in this case, users has one company, company has many users ), the following error appears:

```
[error] #PID<0.327.0> running JsonapiOverhaul.Endpoint terminated
Server: localhost:4000 (http)
Request: GET /api/companies/1
** (exit) an exception was raised:
    ** (UndefinedFunctionError) undefined function: Enum.empty/1
        (elixir) Enum.empty(nil)
        lib/jsonapi/serializer.ex:51: anonymous fn/6 in JSONAPI.Serializer.encode_data/4
        (elixir) lib/enum.ex:1102: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
        lib/jsonapi/serializer.ex:17: JSONAPI.Serializer.serialize/3
        (phoenix) lib/phoenix/view.ex:341: Phoenix.View.render_to_iodata/3
        (phoenix) lib/phoenix/controller.ex:628: Phoenix.Controller.do_render/4
        (jsonapi_overhaul) web/controllers/company_controller.ex:1: JsonapiOverhaul.CompanyController.action/2
        (jsonapi_overhaul) web/controllers/company_controller.ex:1: JsonapiOverhaul.CompanyController.phoenix_controller_pipeline/2
        (jsonapi_overhaul) lib/phoenix/router.ex:255: JsonapiOverhaul.Router.dispatch/2
        (jsonapi_overhaul) web/router.ex:1: JsonapiOverhaul.Router.do_call/2
        (jsonapi_overhaul) lib/jsonapi_overhaul/endpoint.ex:1: JsonapiOverhaul.Endpoint.phoenix_pipeline/1
        (jsonapi_overhaul) lib/plug/debugger.ex:92: JsonapiOverhaul.Endpoint."call (overridable 3)"/2
        (jsonapi_overhaul) lib/phoenix/endpoint/render_errors.ex:34: JsonapiOverhaul.Endpoint.call/2
        (plug) lib/plug/adapters/cowboy/handler.ex:15: Plug.Adapters.Cowboy.Handler.upgrade/4
        (cowboy) src/cowboy_protocol.erl:442: :cowboy_protocol.execute/4
```

This could be related to the fact that on the tests, i am not filling out the relationships.

Thanks!

jeregrine:
[...]

PR #30, opened 2015-12-22
Added tests for #26.

jakerobers:
bye bye bugs. I did a test for when a relationship object is `nil` and `[]`.


jakerobers:
I did a `git checkout` on the commit from before the round of fixes. After that, I did a `git cherry-pick` for the tests that I just wrote.
The tests fail on the codebase from before today. Now they are passing :tada:

jeregrine:
:heart:

Issue #37, opened 2016-04-12
Relationship self links are incorrectly formatted


mpriestman:
I think the format of relationship `self` links in incorrect (although I'm no expert on JSONAPI). In your unit test, you have a `PostView`, which has a `author` relationship to a `UserView`. You define the relationship as such:

`def relationships(), do: [author: {JSONAPISerializerTest.UserView, :include}]`

I would expect the relationship `author` self link to be: `mytype/1/relationships/author`, but instead it is serialised as `mytype/1/relationships/user`. Essentially, I think it's using the _type_ of the relationship at the end, and not the _name_.

The JSONAPI spec page has a very similar example, and there, when they mention a sample `self` link to an "author", which one presumes is of type "user", they specify the relationship as `/articles/1/relationships/author`

PR #38, opened 2016-04-12
Fixed up relationship 'self' links.

mpriestman:
Uses the key of the relationship instead of the type. Fixes issue #37.

Added a unit test that demonstrates the problem as well.


jeregrine:
Thank you so much!

jeregrine:
This has been published!

PR #41, opened 2016-04-25
Fix up the view example

scrogson:


jeregrine:
<3 Thank you so much!

PR #44, opened 2016-05-03
Get QueryParser working again

scrogson:
This is a WIP PR to fix the includes...

There are failing tests for the serializer...but I've gotten this working on a project locally...so I need to figure out the mismatch between the things...

I think we need to remove the default `:include` tuple from the views and move default includes to the query parser as an option.

```elixir
plug JSONAPI.QueryParser,
  sort: [...],
  filter: [...],
  default_includes: [...],
  view: MyApp.PostView
```

**DO NOT MERGE**


jeregrine:
👍

PR #50, opened 2016-05-05
Allow urls to be namespaced

### LINKED ISSUE #35: Custom URL in links?
Is there a way to customize the links that get built? 

My resource is set up in the router.ex like this...

```Elixir
 scope "/api", Handiman do
   pipe_through :api

   resources "/users", UserController, except: [:new, :edit, :create]
   resources "/register", RegistrationController, only: [:create]
   resources "/rounds", RoundController, except: [:new, :edit]
  end
```

so my URLs will look like `/api/rounds` but when the links are created for 'self' for example, `api` is not included and the URL is useless.

scrogson:
This PR fixes #35.

It also allows you to specify the `type` function via passing an option:

```elixir
defmodule PostView do
  use JSONAPI.View, type: "posts", namespace: "/api"
  ...
end
```

### DIFF
--- a/lib/jsonapi/phoenix.ex
+++ b/lib/jsonapi/phoenix.ex
@@ -1,16 +1,23 @@
 defmodule JSONAPI.PhoenixView do
   @moduledoc """
-  This is an optional Phoenix specific module to include. It will give you default render show and index.json methods.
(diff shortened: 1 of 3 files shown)


jeregrine:
💖💕🚀💜

PR #62, opened 2017-06-28
Add support for meta fields

### LINKED ISSUE #56: Allow to set meta
Would be nice to allow to set meta for each resources. Something like this I imagine:

```
# message_view.ex
defmodule Foo.MessageView do
  def meta
    %{ count: Message.count, unread_count: Message.unread_count_for(params[:id]) }
  end
end
```

Not sure how we can pass the "params" in there though. Any thoughts?

doomspork:
@jeregrine / @sebbean / @gottfrois — feedback welcomed and encouraged!

This closes #56.

### DIFF
--- a/lib/jsonapi/serializer.ex
+++ b/lib/jsonapi/serializer.ex
@@ -15,7 +15,7 @@ defmodule JSONAPI.Serializer do
     end
 
     {to_include, encoded_data} = encode_data(view, data, conn, query_includes)
-    
+
     %{
       links: %{},
       data: encoded_data,
@@ -43,6 +43,12 @@ defmodule JSONAPI.Serializer do
         self: view.url_for(data, conn)
       }
     }
+
+    doc =
+      case view.meta(data, conn) do
+        nil -> doc
+        meta -> Map.put(doc, :meta, meta)
+      end
 
     # Handle all the relationships
     Enum.map_reduce(view.relationships(), doc, fn({key, include_view}, acc) ->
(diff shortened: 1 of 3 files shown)


lib/jsonapi/serializer.ex
```diff
@@ -38,6 +38,7 @@ defmodule JSONAPI.Serializer do
       id: view.id(data),
       type: view.type(),
       attributes: view.attributes(data, conn),
+      meta: view.meta(data, conn),
```
jeregrine: I think we should only add meta if view.meta() is not nil? The spec doesn't state if we should send down empty maps or not but if someone isn't setting meta perhaps they'd prefer it not be sent down?

Subsequently we should return nil here. https://github.com/jeregrine/jsonapi/pull/62/files#diff-c44c9da7e2c91d6f0643f4ef73ab636cR88

jeregrine:
Thank you so much for this though! Really appreciate the assistance.

lib/jsonapi/serializer.ex
doomspork: @jeregrine I coded this based on your comments in the original issue:

> have a default view.ex meta function that returns an empty %{} and is overrideable.

You want to deviate from that now?

doomspork:
@jeregrine I've gone ahead and changed this to return `nil` and only include `meta` in the event that the value is _not_ `nil`.

jeregrine:
Thank you!!

doomspork:
You're welcome @jeregrine 👍

PR #63, opened 2017-07-01
Allow for underscores to be converted to dashes.

### LINKED ISSUE #12: Replace underscores for dashs in multi word names
Seems that according to jsonapi spec and ember data jsonapi,

some_thing , should become some-thing.

In the jsonapi.org page, it has the example:
....

```
attributes": {
      "first-name": "Dan",
      "last-name": "Gebhardt",
      "twitter": "dgeb"
    },
```

....

If i manage to learn how to do this, i can do it on the weekend and write some tests.

doomspork:
@jeregrine would love to get your thoughts on this first pass.

This closes #12

### DIFF
--- a/lib/jsonapi/view.ex
+++ b/lib/jsonapi/view.ex
@@ -82,7 +82,15 @@ defmodule JSONAPI.View do
 
       #TODO Figure out the nesting of fields
       def attributes(data, conn) do
-        Map.take(data, fields())
+        underscore_to_dash = Application.get_env(:jsonapi, :underscore_to_dash, false)
+
+        data = Map.take(data, fields())
+
+        if underscore?() do
+          underscore(data)
+        else
+          data
+        end
       end
 
       def meta(_data, _conn), do: nil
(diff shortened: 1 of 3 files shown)


lib/jsonapi/view.ex
```diff
@@ -91,6 +93,23 @@ defmodule JSONAPI.View do
 
       def fields, do: raise "Need to implement fields/0"
 
+      if Keyword.get(unquote(opts), :underscore_to_dash, false) do
```
doomspork: Is there a better option name?

doomspork:
Let me know if there's any changes you'd like to see @jeregrine 😁

lib/jsonapi/view.ex
jeregrine: I think this would be better served as an application configuration since it would MOST likely be something they would use for the whole project and not a single view.

doomspork: That's easy enough, still the same option name @jeregrine?

jeregrine: Ahh year seems good.

doomspork:
@jeregrine take a peek!  I made a few changes:

1) Now it relies on `Application.get_env(:jsonapi, :underscore_to_dash)` for the configuration option.
2) Changed the function handling the underscore from `sanitize/1` to `underscore/1` to better reflect it's job.

jeregrine:
Perfect! Thank you so much!

doomspork:
You're welcome @jeregrine 👍

PR #65, opened 2017-07-10
Correct README for calling serialization without phoenix

bmrsny:
First just wanted to say thanks for this awesome library!

I noticed an issue with the README while working through some different usages. The call to `serialize` was incorrect in the README when you are not using phoenix. 

Just a quick little README correction.


jeregrine:
Thank you!

bmrsny:
NP!

PR #66, opened 2017-07-10
Add Host Configuration option

bmrsny:
This PR adds the ability to add a host configuration that takes president over `conn.host`
Also adds the ability to add a scheme configuration that takes president over `conn.scheme`

Tested within our application and it works.
ex: `links: %{self: "http://www.not_example.com/api/user/535937“}`

### DIFF
--- a/lib/jsonapi/view.ex
+++ b/lib/jsonapi/view.ex
@@ -106,17 +106,16 @@ defmodule JSONAPI.View do
       end
 
       def url_for(data, %Plug.Conn{}=conn) when is_list(data) do
-        "#{Atom.to_string(conn.scheme)}://#{conn.host}#{@namespace}/#{type()}"
+        "#{scheme(conn)}://#{host(conn)}#{@namespace}/#{type()}"
       end
 
       def url_for(data, %Plug.Conn{}=conn) do
-        "#{Atom.to_string(conn.scheme)}://#{conn.host}#{@namespace}/#{type()}/#{id(data)}"
+        "#{scheme(conn)}://#{host(conn)}#{@namespace}/#{type()}/#{id(data)}"
       end
 
       def url_for_rel(data, rel_type, conn) do
         "#{url_for(data, conn)}/relationships/#{rel_type}"
       end
-
 
       if Code.ensure_loaded?(Phoenix) do
         def render("show.json", %{data: data, conn: conn}),
@@ -129,6 +128,10 @@ defmodule JSONAPI.View do
         def render("index.json", %{data: data, conn: conn, params: params}),
           do: show(data, conn, params)
       end
+
+      defp host(conn), do: Application.get_env(:jsonapi, :host, conn.host)
+
+      defp scheme(conn), do: Application.get_env(:jsonapi, :scheme, to_string(conn.scheme))
 
       defoverridable attributes: 2,
                      fields: 0,
--- a/test/jsonapi/view_test.exs
+++ b/test/jsonapi/view_test.exs
@@ -23,7 +23,7 @@ defmodule JSONAPI.ViewTest do
     assert PostView.type == "posts"
   end
 
-  test "url_for/2 with namespace" do
+  test "url_for/2" do
     assert PostView.url_for(nil, nil) == "/api/posts"
     assert PostView.url_for([], nil) == "/api/posts"
     assert PostView.url_for(%{id: 1}, nil) == "/api/posts/1"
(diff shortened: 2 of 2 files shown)


bmrsny:
Couldn't find an elegant way to make this application wide. Any suggestions are appreciated.

Also not a fan of the case statements and open to suggestions to clean this up as well.

lib/jsonapi/view.ex
```diff
@@ -63,12 +63,14 @@ defmodule JSONAPI.View do
   defmacro __using__(opts \\ []) do
     {type, opts} = Keyword.pop(opts, :type)
     {namespace, _opts} = Keyword.pop(opts, :namespace, "")
+    {host, _opts} = Keyword.pop(opts, :host, "")
```
bmrsny: not sure if `nil` is a better option for a default.

ybur-yug: I think `nil` is more appropriate

ybur-yug: But that might mess up some of the other string bits

jeregrine: If you wanted to have it be application wide you could do Application.get_env(:jsonapi, :host)

ybur-yug:
Also, this is opened to attempt to close #66

lib/jsonapi/view.ex
```diff
@@ -94,23 +96,38 @@ defmodule JSONAPI.View do
         do: serialize(__MODULE__, models, conn)
 
       def url_for(nil, nil) do
-        "#{@namespace}/#{type()}"
+        case @host do
```
doomspork: Are all of these case statements really necessary?  This is already inside of a macro, why not only generate the methods that are necessary and avoid the necessary `case`?

lib/jsonapi/view.ex
```diff
@@ -94,23 +96,38 @@ defmodule JSONAPI.View do
...
+          ""   -> "#{@namespace}/#{type()}/#{id(data)}"
+          host -> "#{host}#{@namespace}/#{type()}/#{id(data)}"
+        end
       end
 
       def url_for(data, %Plug.Conn{}=conn) when is_list(data) do
-        "#{Atom.to_string(conn.scheme)}://#{conn.host}#{@namespace}/#{type()}"
+        case @host do
+          ""   -> "#{Atom.to_string(conn.scheme)}://#{conn.host}#{@namespace}/#{type()}"
```
doomspork: If we're going to make the host configurable, shouldn't the scheme be as well?  If you're using SSL termination at your LB your application is going to receive HTTP requests and generate HTTP URLs.

lib/jsonapi/view.ex
```diff
@@ -94,23 +96,38 @@ defmodule JSONAPI.View do
...
-        "#{@namespace}/#{type()}"
+        case @host do
+        ""   -> "#{@namespace}/#{type()}"
+        host -> "#{host}#{@namespace}/#{type()}"
+        end
       end
 
       def url_for(data, nil) when is_list(data) do
-        "#{@namespace}/#{type()}"
+        case @host do
```
doomspork: These intentionally do not have a host because there is no connection, are we should we want to start appending the host to these?  It may have unintended consequences.

test/jsonapi/view_test.exs
```diff
@@ -33,6 +35,14 @@ defmodule JSONAPI.ViewTest do
     assert PostView.url_for_rel(%{id: 1}, "comments", %Plug.Conn{}) == "http://www.example.com/api/posts/1/relationships/comments"
   end
 
+  test "url_for/2 with custom host" do
+    assert CommentView.url_for(nil, nil) == "www.notexample.com/api/comments"
```
doomspork: The `CommentView` module is only included for Phoenix tests, will this bomb out in the event `tags[:compile_phoenix]` is not set?

doomspork:
I think this PR can simplified down quite a bit.  To achieve the same goal, and allow for scheme to be configurable as well, we should only need to change about four lines:

```elixir
      def url_for(data, %Plug.Conn{}=conn) when is_list(data) do
        "#{scheme(conn)}://#{host(conn)}#{@namespace}/#{type()}"
      end

      def url_for(data, %Plug.Conn{}=conn) do
        "#{scheme(conn)}://#{host(conn)}#{@namespace}/#{type()}/#{id(data)}"
      end

      defp host(conn), do: Application.get_env(:jsonapi, :host) || conn.host

      defp scheme(conn), do: Application.get_env(:jsonapi, :scheme) || to_string(conn.scheme)
```

jeregrine:
@bmrsny Thank you so much for the contribution. I think we should go with something similar to what @doomspork did in his branch here. https://github.com/doomspork/jsonapi/tree/feature/configurable-host-and-scheme could you make similar changes?

lib/jsonapi/view.ex
bmrsny: You know Im not completely sure. I had noticed that namespace was being used and thought putting the host made sense at the time, however I do agree that if there is no connection there isn't a need for a host and possibly a namespace as well. Is there a preferred behavior we want?

doomspork: I would keep the existing library behavior, that's always the safest bet.

bmrsny:
Updated the PR with changes that reflect @doomspork 's branch. I had to add an `Application.delete_env` to the test because the test would fail intermittently depending on the order that they ran in. 

I did not incorporate the underscore_to_dash portions. I felt like it was out of scope with dealing with  `url_for` and was dealing with attributes data. I can include it if we want to move forward with that functionality, however.

Let me know your thoughts. Thanks all for your help and the opportunity to contribute to this awesome library!

lib/jsonapi/view.ex
```diff
@@ -130,6 +129,10 @@ defmodule JSONAPI.View do
           do: show(data, conn, params)
       end
 
+      defp host(conn), do: Application.get_env(:jsonapi, :host) || conn.host
```
doomspork: Can you update these to reflect the comment I left?  I made some changes there not in the branch: https://github.com/jeregrine/jsonapi/pull/66#issuecomment-314489047

test/jsonapi/view_test.exs
```diff
@@ -33,6 +33,26 @@ defmodule JSONAPI.ViewTest do
...
+    Application.put_env(:jsonapi, :host, "www.otherhost.com")
+    assert PostView.url_for([], %Plug.Conn{}) == "http://www.otherhost.com/api/posts"
+    assert PostView.url_for(%{id: 1}, %Plug.Conn{}) == "http://www.otherhost.com/api/posts/1"
+    assert PostView.url_for_rel([], "comments", %Plug.Conn{}) == "http://www.otherhost.com/api/posts/relationships/comments"
+    assert PostView.url_for_rel(%{id: 1}, "comments", %Plug.Conn{}) == "http://www.otherhost.com/api/posts/1/relationships/comments"
+    Application.delete_env(:jsonapi, :host)
+  end
+
+  test "url_for/2 with configured scheme" do
+    Application.put_env(:jsonapi, :host, "otherhost.com")
```
doomspork: If you take a peek at my branch you'll see I consolidated these tests together for this very reason.  Unfortunately since the tests run currently, this can still cause an issue just less frequently.  Can we clean this up a bit?

doomspork:
These changes look good to me @bmrsny, thanks for the contribution!  I'll defer to @jeregrine for the merge 👍

jeregrine:
Thank you so muh ❤️💖

PR #70, opened 2017-07-12
Add hidden/0 to view

### LINKED ISSUE #68: Hidden fields
@jeregrine curious to get your thoughts on whether this is a feature we want or something I should keep in my application.  There are a few instances in which we need to include a field in our view so it can be posted or patched but we don't want it included in the response (think passwords).  For this I've implemented a `hidden/0` that works like `fields/0` but is used to hide certain values from the response.

Is this something you'd be open to considering?  If so I'll open a PR.

doomspork:
This closes #68.  I went with `hidden/0` but I'm happy to change this to the tuple format instead.

### DIFF
--- a/lib/jsonapi/view.ex
+++ b/lib/jsonapi/view.ex
@@ -82,9 +82,7 @@ defmodule JSONAPI.View do
 
       #TODO Figure out the nesting of fields
       def attributes(data, conn) do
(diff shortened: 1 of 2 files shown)


jeregrine:
🎉

doomspork:
Thanks @jeregrine!

PR #71, opened 2017-07-19
Document configuration options in README and view module

ybur-yug:
Hey @jeregrine!

Just thought since there were the new configuration options added in 0.5.0 that I would throw a configuration section into the README so it was more clear and also make a mention in the doc for the view module.

These arent exactly huge, but I thought it might make things a little less bumpy for people who might not have dove into the code to realize where there are configurations being thrown in.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,18 @@ when Ecto gets more complex field selection support we will go further to only q
 
 You will need to handle filtering yourself, the filter is just a map with key=value.
 
+## Configuration
+By default host and scheme are pulled from the provided conn but can be overridden via configuration like so:
+
+```elixir
+config :jsonapi,
+  host: "www.someotherhost.com",
+  scheme: "https",
+  underscore_to_dash: true
+```
+
+Additionally, JSONAPI now recommends the use of dashes (`-`) in place of underscore (`_`) as a word separator. Enabling this change is easy with the `underscore_to_dash` option, which handles the conversion for you.
+
 ## Other
 
 - Feel free to make PR's. I will do my best to respond within a day or two.
--- a/lib/jsonapi/view.ex
+++ b/lib/jsonapi/view.ex
@@ -59,6 +59,16 @@ defmodule JSONAPI.View do
   If you always want to include a relationship. First make sure its always preloaded
   and then use the `[user: {UserView, :include}]` syntax in your `includes` function. This tells
   the serializer to *always* include if its loaded.
+
+  ## Options
+    * `:host` (binary) - Allows the `host` to be overrided for generated URLs.  Defaults to `host` of the supplied `conn`.
+    
+    * `:scheme` (atom) - Enables configuration of the HTTP scheme for generated URLS.  Defaults to `scheme` from the provided `conn`.
+
+    * `:underscore_to_dash` (boolean) - Use dash (`-`) as the word separated for JSON in place of underscore (`_`) per the JSONAPI spec [recommendations](http://jsonapi.org/recommendations/).  Defaults to `false`.
+
+  The default behaviour for `host` and `scheme` is to derive it from the `conn` provided, while the
+  default style for presentation in names is to be underscored and not dashed.
   """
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -60,6 +60,22 @@ when Ecto gets more complex field selection support we will go further to only q
...
+
+There are several options available through configuration.
+These include switching from underscores to dashes in serialization, customizing the host name, and customizing the URL scheme.
+
+Example:
+
+```elixir
+config :jsonapi,
+  host: "www.someotherhost.com",
+  scheme: "ftp",
```
doomspork: I think `https` would be a better example here.  JSONAPI and `ftp` don't typically work together.

lib/jsonapi/view.ex
```diff
@@ -59,6 +59,17 @@ defmodule JSONAPI.View do
   If you always want to include a relationship. First make sure its always preloaded
   and then use the `[user: {UserView, :include}]` syntax in your `includes` function. This tells
   the serializer to *always* include if its loaded.
+
+  There are several options available through configuration. These include switching from underscores
```
doomspork: Do we need to copy + paste the README doc into here?  I think we can do without.

doomspork:
With a few changes this should be good to merge, though there's a large change to the views coming that will require more in-depth documentation.

README.md
```diff
@@ -60,6 +60,22 @@ when Ecto gets more complex field selection support we will go further to only q
 
 You will need to handle filtering yourself, the filter is just a map with key=value.
 
+## Configuration
+
+There are several options available through configuration.
```
doomspork: Thinking about this some more I think the defaults should be mentioned first and not as a side note as the end.  How about something like this:

------

By default `host` and `scheme` are pulled from the provided `conn` but can be overridden via configuration:

```elixir
config :jsonapi,
  host: "www.someotherhost.com",
  scheme: "https",
  underscore_to_dash: true
```

JSONAPI now recommends the use of dashes (`-`) in place of underscore (`_`) as a word separator, supporting this change is easy with the `underscore_to_dash` option which handles the conversion for you.

------

I think it _might_ be worth mentioning exactly _why_ we have an option called `underscore_to_dash` since that won't be immediately clear to others.

lib/jsonapi/view.ex
ybur-yug: My thought was if browsing the docs in-terminal the README wouldnt be there if you're doing `h JSONAPI.View`, and this makes sure its available to both casual browsers on hex.pm/github and also those who use the IEx style of doc-reading.

README.md
ybur-yug: Sounds good to me. Will make this change + add a note on the reasoning when I have a bit of time later today.

lib/jsonapi/view.ex
doomspork: IMO inline documentation should be more of just that: documentation.  A code example does not really inform what these do.  I think if we intend to document options in the code, we should follow a similar pattern to Elixir and other libraries: https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/string.ex#L297

README.md
doomspork: @ybur-yug I think the last sentence I tossed in there should be sufficient for the reasoning so long as we include a link to the spec: http://jsonapi.org/recommendations/

lib/jsonapi/view.ex
doomspork:
```elixir
## Options

  * `:host` (binary) - Allows the `host` to be overrided for generated URLs.  Defaults to `host` of the supplied `conn`.
  
  * `:scheme` (atom) - Enables configuration of the HTTP scheme for generated URLS.  Defaults to `scheme` from the provided `conn`.

  * `:underscore_to_dash` (boolean) - Use dash (`-`) as the word separated for JSON in place of underscore (`_`) per the JSONAPI spec [recommendations](http://jsonapi.org/recommendations/).  Defaults to `false`

```

PR #73, opened 2017-08-03
Put correct render call in README.md

nickgartmann:
Phoenix `render/4` requires the order of `conn`, `ViewModule`, `action`, `data` and the current readme has the view module and action swapped.


jeregrine:
Thanks!

PR #77, opened 2017-10-09
Dashes in relationships key

mitchellhenke:
Not small refactor of `underscore_to_dash` to make relationship keys also have that functionality.

if `underscore_to_dash` is set, relationship keys will also now have their underscores replaced by dashes.

```elixir
%{
  data: %{
    attributes: %{
      full_description: "description"
    }
  },
  relationships: %{
    best_comments: []
  }
}
```

We'll get:

```elixir
%{
  data: %{
    attributes: %{
      "full-description": "description"
    }
  },
  relationships: %{
    "best-comments": []
  }
}
```

### DIFF
--- a/lib/jsonapi/serializer.ex
+++ b/lib/jsonapi/serializer.ex
@@ -4,6 +4,7 @@ defmodule JSONAPI.Serializer do
   """
 
   import JSONAPI.Ecto, only: [assoc_loaded?: 1]
+  alias JSONAPI.Utils.Underscore
 
   @doc """
   Takes a view, data and a optional plug connection and returns a fully JSONAPI Serialized document.
@@ -41,7 +42,7 @@ defmodule JSONAPI.Serializer do
     doc = %{
       id: view.id(data),
       type: view.type(),
-      attributes: view.attributes(data, conn),
+      attributes: underscore(view.attributes(data, conn)),
       relationships: %{},
       links: %{
         self: view.url_for(data, conn)
@@ -75,7 +76,7 @@ defmodule JSONAPI.Serializer do
     # Build the relationship url
     rel_url = view.url_for_rel(data, key, conn)
     # Build the relationship
-    acc = put_in(acc, [:relationships, key], encode_relation(only_rel_view, rel_data, rel_url, conn))
+    acc = put_in(acc, [:relationships, underscore(key)], encode_relation(only_rel_view, rel_data, rel_url, conn))
 
     valid_include_view = include_view(valid_includes, key)
 
@@ -163,4 +164,12 @@ defmodule JSONAPI.Serializer do
 
   def get_view({view, :include}), do: view
   def get_view(view), do: view
+
+  def underscore(data) do
+    if Underscore.underscore?() do
+      Underscore.underscore(data)
+    else
+      data
+    end
+  end
 end
--- a/lib/jsonapi/utils/underscore.ex
+++ b/lib/jsonapi/utils/underscore.ex
@@ -1,0 +1,31 @@
+defmodule JSONAPI.Utils.Underscore do
+  @moduledoc """
+  Helpers for replacing underscores with dashes.
+  """
+
+  def underscore?, do: Application.get_env(:jsonapi, :underscore_to_dash, false)
+
+  def underscore(value) when is_atom(value) do
+    value
+    |> to_string
+    |> underscore
+  end
+
+  def underscore(value) when is_binary(value) do
+    String.replace(value, "_", "-")
+  end
+
+  def underscore(value) when is_map(value) do
+    value
+    |> Enum.map(&underscore/1)
+    |> Enum.into(%{})
+  end
+
+  def underscore({key, value}) do
+    if is_map(value) do
+      {underscore(key), underscore(value)}
+    else
+      {underscore(key), value}
+    end
+  end
+end
--- a/test/serializer_test.exs
+++ b/test/serializer_test.exs
@@ -1,23 +1,24 @@
 defmodule JSONAPISerializerTest do
-  use ExUnit.Case
+  use ExUnit.Case, async: false
   alias JSONAPI.Serializer
 
   defmodule PostView do
     use JSONAPI.View
 
-    def fields, do: [:text, :body]
+    def fields, do: [:text, :body, :full_description]
     def meta(data, _conn), do: %{meta_text: "meta_#{data[:text]}"}
     def type, do: "mytype"
     def relationships do
       [author: {JSONAPISerializerTest.UserView, :include},
-       comments: {JSONAPISerializerTest.CommentView, :include}]
+       best_comments: {JSONAPISerializerTest.CommentView, :include}
+      ]
     end
   end
 
   defmodule UserView do
     use JSONAPI.View
 
-    def fields, do: [:username]
+    def fields, do: [:username, :first_name, :last_name]
     def type, do: "user"
     def relationships, do: []
   end
@@ -46,7 +47,7 @@ defmodule JSONAPISerializerTest do
       text: "Hello",
       body: "Hello world",
       author: %{ id: 2, username: "jason"},
-      comments: [
+      best_comments: [
         %{ id: 5, text: "greatest comment ever", user: %{id: 4, username: "jack"}},
         %{ id: 6, text: "not so great", user: %{id: 2, username: "jason"}}
       ]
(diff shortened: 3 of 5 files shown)


lib/jsonapi/utils/underscore.ex
```diff
@@ -0,0 +1,17 @@
...
+  def underscore?, do: Application.get_env(:jsonapi, :underscore_to_dash, false)
+
+  # TODO Figure out the nesting of fields
+  def underscore(data) do
+    data
+    |> Enum.map(fn {key, value} -> {underscore_key(key), value} end)
+    |> Enum.into(%{})
+  end
+
+  def underscore_key(key) do
```
doomspork: Perhaps we could refactor out the explicit `underscore_key/1` and rely on pattern matching instead, that might help us solve nesting:

```elixir
defmodule JSONAPI.Utils.Underscore do
  def underscore?, do: Application.get_env(:jsonapi, :underscore_to_dash, false)

  def underscore(value) when is_atom(value) do
    value
    |> to_string
    |> underscore
  end

  def underscore(value) when is_bitstring(value) do
    value
    |> String.replace("_", "-")
    |> String.to_atom
  end

  def underscore(value) when is_map(value) do
    value
    |> Enum.map(&underscore/1)
    |> Enum.into(%{})
  end

  def underscore({key, value}) do
    {underscore(key), underscore(value)}
  end
end
```

test/serializer_test.exs
```diff
@@ -1,23 +1,24 @@
 defmodule JSONAPISerializerTest do
-  use ExUnit.Case
+  use ExUnit.Case, async: false
```
doomspork: It's a shame we have to make this non-async.  I'm wondering if that suggests we should re-think this approach a bit?

lib/jsonapi/serializer.ex
```diff
@@ -163,4 +163,20 @@ defmodule JSONAPI.Serializer do
 
   def get_view({view, :include}), do: view
   def get_view(view), do: view
+
+  def maybe_dasherize_key(data) do
```
doomspork: Can we DRY these methods up at all?  If we remove `underscore_key/1` in favor of pattern matching in the Util module we might be able to.

I think we should change `dasherize` to `underscore` so that the code matches up to the configuration it's dealing with.  Also, is the maybe prefix necessary?

doomspork:
@mitchellhenke looks like we're missing some things Credo requires.  Would you mind squashing your commits? 😁

test/serializer_test.exs
mitchellhenke: It's required because of the environment modification.  Best option is probably an argument to one of the functions, but I'm not sure which is the right option.

lib/jsonapi/serializer.ex
mitchellhenke: I used `dasherize` as it felt like it better described what's happening to the data, and `maybe_` is a pattern I've seen used to indicate something may or may not happen.  I've replaced `maybe_dasherize` with `underscore` for consistency with the configuration.

lib/jsonapi/utils/underscore.ex
mitchellhenke: Sounds good!  I had to change the final clause as underscoring map string values caused some issues.

mitchellhenke:
@doomspork I've updated per your comments with the exception of the async tests, as I'm not sure how to best work around that.

lib/jsonapi/utils/underscore.ex
```diff
@@ -0,0 +1,33 @@
...
+
+  def underscore?, do: Application.get_env(:jsonapi, :underscore_to_dash, false)
+
+  def underscore(value) when is_atom(value) do
+    value
+    |> to_string
+    |> underscore
+  end
+
+  def underscore(value) when is_bitstring(value) do
```
jeregrine: this should be `is_binary`

lib/jsonapi/utils/underscore.ex
```diff
@@ -0,0 +1,33 @@
...
+    value
+    |> String.replace("_", "-")
+    |> String.to_atom
```
jeregrine: And Actually I don't think we need to "to_atom" they all turn to strings in the end.

test/serializer_test.exs
jeregrine: The tests are fast enough that its largely a non-issue anyways :P

doomspork: I have no doubt they are!  I just always stop to reconsider the approach when tests starting have to be sequential, tends to indicate some type of smell (at least in my experience).

lib/jsonapi/serializer.ex
```diff
@@ -163,4 +163,12 @@ defmodule JSONAPI.Serializer do
 
   def get_view({view, :include}), do: view
   def get_view(view), do: view
+
+  def underscore(data) do
+    if JSONAPI.Utils.Underscore.underscore?() do
```
doomspork: Should this be aliased?

test/serializer_test.exs
mitchellhenke: @doomspork do you think an `opts` arg should be added to `JSONAPI.Serializer.serialize`?

lib/jsonapi/utils/underscore.ex
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  def underscore(value) when is_map(value) do
+    value
+    |> Enum.map(&underscore/1)
+    |> Enum.into(%{})
+  end
+
+  def underscore({key, value}) do
+    if is_map(value) do
```
doomspork: I don't think this check is necessary, do you?  When you pass value through `underscore/1` the previously defined functions should capture the map or the binary.  I don't think there's any reason to _not_ run `value` through `underscore/1`

lib/jsonapi/utils/underscore.ex
```diff
@@ -0,0 +1,32 @@
...
+
+  def underscore(value) when is_atom(value) do
+    value
+    |> to_string
+    |> underscore
+  end
+
+  def underscore(value) when is_binary(value) do
+    value
+    |> String.replace("_", "-")
```
doomspork: We don't use pipes when there is a single input.

lib/jsonapi/utils/underscore.ex
doomspork: Out of curiosity @jeregrine, why do you prefer `is_binary` over `is_bitstring`?  I've seen the reverse suggestions previously (on other projects).

test/serializer_test.exs
doomspork: That's a great question/idea @mitchellhenke!  I haven't thought about it. 
 Do you think it would make it more difficult for folks to leverage the library and modules?

I don't think there's too many cases where folks use `serialize/3` directly, do you?

On a side note: I need the 🤔  for a reaction!

lib/jsonapi/utils/underscore.ex
mitchellhenke: I think it is with the current flow.  Recursively calling underscore on all of the values means it will be called on regular values, e.g.

```
%{
  id: 1,
  my_text: "my_text"
  # ...
}
```

gets serialized to

```
%{
  id: 1,
  my-text: "my-text"
}
```

The two options I see are to either not do it recursively, or do it recursively and avoid calling `underscore` on non-map values.

lib/jsonapi/utils/underscore.ex
mitchellhenke: fixed! 🙂

lib/jsonapi/serializer.ex
mitchellhenke: fixed!

lib/jsonapi/utils/underscore.ex
doomspork: You're right @mitchellhenke, we should leave it as-is 👍

doomspork:
Thanks @mitchellhenke, looks good 👍

test/serializer_test.exs
mitchellhenke: @doomspork I'm not sure.  The `conn` argument could potentially go in an `opts` as well, but it changes the API of anyone using it directly.

mitchellhenke:
Thanks!  And thank you for your patience and help in reviewing it!

lib/jsonapi/utils/underscore.ex
jeregrine: A bit string isn't QUITE a string or binary, as well as the docs use is_binary to select a binary string.

https://elixir-lang.org/getting-started/binaries-strings-and-char-lists.html

PR #79, opened 2017-11-02
Support Limit and Offset as part of the opts map

StareIntoTheBeard:


lib/jsonapi/query_parser.ex
```diff
@@ -65,6 +65,8 @@ defmodule JSONAPI.QueryParser do
     |> parse_include(Map.get(query_params, "include", ""))
     |> parse_filter(Map.get(query_params, "filter", %{}))
     |> parse_sort(Map.get(query_params, "sort", ""))
+    |> Map.put(:limit, Map.get(query_params, "limit", ""))
```
doomspork: If we're going to add these, we should add all the query params JSON API supports for pagination:

+ limit
+ offset
+ size
+ page
+ cursor

These should probably return integers and not strings with `nil` being a default and not `""`, don't you think?

We can probably refactor this whole pipe when we do add the aforementioned params to be a big cleaner and avoid the nested function calls.

StareIntoTheBeard: Yeah, I'll buy that -- side note is there a native ecto syntax for `cursor`?

StareIntoTheBeard: Main question being, what values are we expecting that to come across with?

jeregrine:
@StareIntoTheBeard @doomspork SO the jsonapi spec is very clear on how its agnostic about paging. It says that the "page"keyword is reserved for paging and its to be used like `page[PAGING_TYPE]` http://jsonapi.org/format/#fetching-pagination so if anything I'd support adding the "page" keyword and nothing else.

I'd mostly suggest just grabbing the page keyword from the regular query params and handling it yourself.

I do appreciate the PR though. If you have anything else you'd like to see lemme know.

StareIntoTheBeard:
@jeregrine The point was I wanted it in the same struct with the rest of the JSON API params. I could have easily handed down more query params to Ecto but the spec does handle pagination (in the way you've said, thanks for pointing that out) but I'd like it in the same assigns struct... Made some changes and will reopen and welcome comments.

PR #80, opened 2017-11-03
Limit offset support

StareIntoTheBeard:


StareIntoTheBeard:
Reopening here as I'd like to see this be apart of the opts struct w/ all other JSONAPI compliant query features.

lib/jsonapi/query_parser.ex
```diff
@@ -194,4 +201,6 @@ defmodule JSONAPI.QueryParser do
     view = Keyword.fetch!(opts, :view)
     struct(JSONAPI.Config, opts: opts, view: view)
   end
+
+  defp atomize_map(map), do: for {key, val} <- map, into: %{}, do: {String.to_atom(key), val}
```
jeregrine: We do not want to do this. Atoms are not garbage collected. The reason I did this with map.get was to only pull off the keys I needed before setting them in the struct.

StareIntoTheBeard: Sure, but, correct me if i'm wrong, we're talking about 4 additional atoms here (atoms are unique so `:page` counts once).  
Reference:
https://medium.com/everydayhero-engineering/elixir-atoms-3b0f06bc47ef

I'm just atomizing the keys in the map for ease of converting to a struct.

StareIntoTheBeard: Actually, I suppose I see the issue since this doesn't limit query parameters prior to atomization. Let me make a change to this.

StareIntoTheBeard:
@jeregrine updates such that it only atomizes keys that are apart of the struct.

lib/jsonapi/query_parser.ex
```diff
@@ -194,4 +198,16 @@ defmodule JSONAPI.QueryParser do
     view = Keyword.fetch!(opts, :view)
     struct(JSONAPI.Config, opts: opts, view: view)
   end
+
+  defp struct_from_map(params, struct) do
+    struct_keys = Map.keys(struct) |> Enum.filter(fn x -> x != :__struct__ end)
```
jeregrine: https://hexdocs.pm/elixir/Map.html#from_struct/1 Map.from_struct will accomplish this.

lib/jsonapi/query_parser.ex
```diff
@@ -194,4 +198,16 @@ defmodule JSONAPI.QueryParser do
...
+    processed_map =
+    for struct_key <- struct_keys, into: %{} do
```
jeregrine: To save a step
```
for {key, _ } <- Map.from_struct(keys), into: %{} do
 ..
end
```

StareIntoTheBeard:
Good call, will update.

StareIntoTheBeard:
@jeregrine @doomspork LMK if you think this is reasonable to merge now.

lib/jsonapi/query_parser.ex
```diff
@@ -138,7 +150,7 @@ defmodule JSONAPI.QueryParser do
   def build_sort("", field), do: [asc: field]
   def build_sort("-", field), do: [desc: field]
 
-  def parse_include(config, ""), do: config
+  def parse_include(config, []), do: config
   def parse_include(%Config{} = config, include_str) do
     includes = handle_include(include_str, config)
     Map.put(config, :includes, includes)
```
StareIntoTheBeard: @jeregrine Seems like this should be `include` rather than includes but I'm hesitant to change it so it doesn't break other people's code. Thoughts?

doomspork:
@StareIntoTheBeard could you add test coverage?

StareIntoTheBeard:
@jeregrine @doomspork test coverage added.

lib/jsonapi/query_parser.ex
jeregrine: @StareIntoTheBeard lets support both then? Take include and includes, merge them and return :include and :includes with the same value?

This way we can be backwards and forward compatible?

doomspork: We could raise a deprecation warning for `includes`?

StareIntoTheBeard: @doomspork yeah that's a great idea.

lib/jsonapi/query_parser.ex
```diff
@@ -138,10 +151,14 @@ defmodule JSONAPI.QueryParser do
...
 
-  def parse_include(config, ""), do: config
+  def parse_include(config, []), do: config
   def parse_include(%Config{} = config, include_str) do
     includes = handle_include(include_str, config)
+
+    Deprecation.warn(:includes)
+
     Map.put(config, :includes, includes)
+    |> Map.put(:include, includes)
```
doomspork: Can you update this:

```elixir
config
|> Map.put(:includes, includes)
|> Map.put(:include, includes)
```

test/jsonapi_query_parser_test.exs
```diff
@@ -67,6 +67,13 @@ defmodule JSONAPI.QueryParserTest do
     assert parse_include(config, "comments.user").includes == [comments: :user]
   end
 
+  test "parse_include/2 returns a map with duplicate values for include and includes for compatibility" do
+    include_list = struct(Config, view: MyView)
+    |> parse_include("comments.user")
```
doomspork:
```elixir
include_list = 
  Config
  |> struct(view: MyView)
  |> parse_include("comments.user")
```

StareIntoTheBeard: @doomspork updated

jeregrine:
Thanks!! <3

StareIntoTheBeard:
Thank you! :)

PR #82, opened 2017-11-05
Optional data - links

snewcomer:
- [x] configure `remove_links`.  Set as default `false` to maintain backwards compat.
- [x] ensure `included` not merged into payload if not configured.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,10 +87,13 @@ By default host and scheme are pulled from the provided conn but can be overridd
 config :jsonapi,
   host: "www.someotherhost.com",
   scheme: "https",
-  underscore_to_dash: true
+  underscore_to_dash: true,
+  remove_links: true
 ```
 
 Additionally, JSONAPI now recommends the use of dashes (`-`) in place of underscore (`_`) as a word separator. Enabling this change is easy with the `underscore_to_dash` option, which handles the conversion for you.
+
+`links` data can optionally be removed from the payload via the configuration above.
 
 ## Other
 
--- a/README.md
+++ b/README.md
@@ -88,12 +88,14 @@ config :jsonapi,
   host: "www.someotherhost.com",
   scheme: "https",
   underscore_to_dash: true,
-  remove_links: true
+  remove_links: false
 ```
 
-Additionally, JSONAPI now recommends the use of dashes (`-`) in place of underscore (`_`) as a word separator. Enabling this change is easy with the `underscore_to_dash` option, which handles the conversion for you.
+* **underscore_to_dash**
+Additionally, JSONAPI now recommends the use of dashes (`-`) in place of underscore (`_`) as a word separator. Enabling this change is easy with the `underscore_to_dash` option, which handles the conversion for you.  Defaults to `false`.
 
-`links` data can optionally be removed from the payload via the configuration above.
+* **remove_links**
+`links` data can optionally be removed from the payload via setting the configuration above to `true`.  Defaults to `false`.
 
 ## Other
 
--- a/README.md
+++ b/README.md
@@ -92,9 +92,11 @@ config :jsonapi,
 ```
 
 * **underscore_to_dash**
+
 Additionally, JSONAPI now recommends the use of dashes (`-`) in place of underscore (`_`) as a word separator. Enabling this change is easy with the `underscore_to_dash` option, which handles the conversion for you.  Defaults to `false`.
 
 * **remove_links**
+
 `links` data can optionally be removed from the payload via setting the configuration above to `true`.  Defaults to `false`.
 
 ## Other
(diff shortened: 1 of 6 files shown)


lib/jsonapi/serializer.ex
```diff
@@ -108,16 +106,25 @@ defmodule JSONAPI.Serializer do
...
-      links: %{
-        self: rel_url,
-        related: rel_view.url_for(rel_data, conn)
-      },
       data: encode_rel_data(rel_view, rel_data)
     }
+    |> merge_related_links(info, Application.get_env(:jsonapi, :remove_links, false))
   end
 
+  defp merge_links(doc, data, view, conn, _remove_links = false) do
```
snewcomer: set the default to `false` to maintain backwards compat.

snewcomer:
I also removed the `included` data if not configured at view level as a second commit.  Lmk what you think!

snewcomer:
Updated branch with master.

README.md
```diff
@@ -87,11 +87,14 @@ By default host and scheme are pulled from the provided conn but can be overridd
 config :jsonapi,
   host: "www.someotherhost.com",
   scheme: "https",
-  underscore_to_dash: true
+  underscore_to_dash: true,
+  remove_links: true
```
jeregrine: set this to false in documentation.

README.md
```diff
@@ -87,11 +87,14 @@ By default host and scheme are pulled from the provided conn but can be overridd
...
 ```
 
 Additionally, JSONAPI now recommends the use of dashes (`-`) in place of underscore (`_`) as a word separator. Enabling this change is easy with the `underscore_to_dash` option, which handles the conversion for you.
 
+`links` data can optionally be removed from the payload via the configuration above.
```
jeregrine: This is a little over the scope of your PR but IF you are willing we should create a "Configuration" section which documents each key and its default. I can create a separate issue too.

snewcomer: @jeregrine So one thing I noticed is this is already in the `Configuration` section.  Let me know what you think about the most recent commit: https://github.com/jeregrine/jsonapi/pull/82/files#diff-04c6e90faac2675aa89e2176d2eec7d8R94

jeregrine:
@snewcomer ugh brain fart, thank you!

jeregrine:
@snewcomer should also fix the credo complaints <3 and then I'll merge this. I'm planning on cutting a new release tomorrow evening.

lib/jsonapi/serializer.ex
```diff
@@ -98,26 +96,35 @@ defmodule JSONAPI.Serializer do
...
-        related: rel_view.url_for(rel_data, conn)
-      },
       data: encode_rel_data(rel_view, rel_data)
     }
+    |> merge_related_links(info, Application.get_env(:jsonapi, :remove_links, false))
   end
 
+  defp merge_links(doc, data, view, conn, false = _remove_links) do
+    doc
+    |> Map.merge(%{links: %{self: view.url_for(data, conn)}})
```
doomspork: This doesn't need to be a pipe: https://github.com/christopheradams/elixir_style_guide#avoid-single-pipelines

lib/jsonapi/serializer.ex
```diff
@@ -98,26 +96,35 @@ defmodule JSONAPI.Serializer do
...
+  end
+  defp merge_links(doc, _data, _view, _conn, _remove_links), do: doc
+
+  defp merge_related_links(encoded_data, {rel_view, rel_data, rel_url, conn}, false = _remove_links) do
+    encoded_data
+    |> Map.merge(%{links: %{self: rel_url, related: rel_view.url_for(rel_data, conn)}})
```
doomspork: Same here: https://github.com/christopheradams/elixir_style_guide#avoid-single-pipelines

lib/jsonapi/serializer.ex
```diff
@@ -98,26 +96,35 @@ defmodule JSONAPI.Serializer do
...
-      links: %{
-        self: rel_url,
-        related: rel_view.url_for(rel_data, conn)
-      },
       data: encode_rel_data(rel_view, rel_data)
     }
+    |> merge_related_links(info, Application.get_env(:jsonapi, :remove_links, false))
   end
 
+  defp merge_links(doc, data, view, conn, false = _remove_links) do
```
doomspork: Please remove the `= _remove_links`, it's not necessary here 😁

lib/jsonapi/serializer.ex
```diff
@@ -43,11 +43,9 @@ defmodule JSONAPI.Serializer do
       id: view.id(data),
       type: view.type(),
       attributes: underscore(view.attributes(data, conn)),
-      relationships: %{},
-      links: %{
-        self: view.url_for(data, conn)
-      }
+      relationships: %{}
     }
+    |> merge_links(data, view, conn, Application.get_env(:jsonapi, :remove_links, false))
```
doomspork: This is a little funky.  Do you think we could capture the above map and add it as the first argument rather than using a pipe?  For one, we don't need pipes when there's a single pipe and it's a bit different to see a pipe setup like this.

snewcomer:
@doomspork fixed!  Possible to squash in the github UI when you merge?  If I do squash, I think I'll lose your comments.  Lmk if that is not the case and I'll squash!

jeregrine:
Thanks! @snewcomer!

PR #87, opened 2017-11-27
Fix nesting for includes

jlvallelonga:
fixes nested includes. query strings like this should now include data as expected:
`?include=user.company.industry.tags`

### DIFF
--- a/lib/jsonapi/serializer.ex
+++ b/lib/jsonapi/serializer.ex
@@ -82,8 +82,8 @@ defmodule JSONAPI.Serializer do
 
     if {rel_view, :include} == valid_include_view && is_data_loaded?(rel_data) do
       rel_query_includes =
-        if is_list(query_includes) do
-          Keyword.get(query_includes, key, [])
+        if Keyword.keyword?(query_includes) do
+          Keyword.take(query_includes, [key]) |> Keyword.values
         else
           []
         end
@@ -163,6 +163,12 @@ defmodule JSONAPI.Serializer do
   end
   defp handle_include({base, :include}, child, acc) do
     handle_include(base.relationships(), child, acc)
+  end
+  defp handle_include(base, children, acc) when is_list(children) do
+    Enum.reduce(children, acc, fn child, acc ->
+      base = if is_list(base), do: base, else: base.relationships()
+      handle_include(base, child, acc)
+    end)
   end
   defp handle_include(base, child, acc) do
     view = if is_list(base), do: Keyword.get(base, child), else: base
--- a/lib/jsonapi/serializer.ex
+++ b/lib/jsonapi/serializer.ex
@@ -83,13 +83,10 @@ defmodule JSONAPI.Serializer do
     if {rel_view, :include} == valid_include_view && is_data_loaded?(rel_data) do
       rel_query_includes =
         if is_list(query_includes) do
-          Enum.map(query_includes, fn include ->
-            case include do
-              {key, value} -> value
-              _ -> nil
-            end
+          Enum.reduce(query_includes, [], fn
+            {^key, value}, acc -> acc ++ [value]
+            _, acc -> acc
           end)
-          |> Enum.reject(&is_nil/1)
           |> List.flatten
         else
           []
--- a/lib/jsonapi/serializer.ex
+++ b/lib/jsonapi/serializer.ex
@@ -150,28 +150,23 @@ defmodule JSONAPI.Serializer do
   end
 
   defp get_includes(view, query_includes) do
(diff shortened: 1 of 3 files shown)


jeregrine:
cc @doomspork if I could get a second set of eyes on this

doomspork:
@jeregrine I'll take a peek today 👍

lib/jsonapi/serializer.ex
```diff
@@ -83,7 +83,14 @@ defmodule JSONAPI.Serializer do
     if {rel_view, :include} == valid_include_view && is_data_loaded?(rel_data) do
       rel_query_includes =
         if is_list(query_includes) do
-          Keyword.get(query_includes, key, [])
+          Enum.map(query_includes, fn include ->
```
doomspork: Could we refactor this entire pipechain into a `Enum.reduce/3`?

```elixir
Enum.reduce(query_includes, [], fn 
  {_key, value}, acc -> acc ++ [value]
  _, acc -> acc
end)

```

lib/jsonapi/serializer.ex
```diff
@@ -144,29 +152,31 @@ defmodule JSONAPI.Serializer do
...
-  defp get_includes(view, []), do: view.relationships()
-  defp get_includes(view, query_includes) when is_list(query_includes) do
-    base = view.relationships()
-    Enum.reduce(query_includes, [], &(handle_include(base, &1, &2)))
-  end
-  defp get_includes(view, include) do
-    base = view.relationships()
-    Keyword.get(base, include)
+  defp get_includes(view, query_includes) do
+    get_default_includes(view) ++ get_query_includes(view, query_includes)
```
doomspork: Let's avoid pipes for single function calls.

```elixir
includes = get_default_includes(view) ++ get_query_includes(view, query_includes)
Enum.uniq(includes)
```

lib/jsonapi/serializer.ex
```diff
@@ -144,29 +152,31 @@ defmodule JSONAPI.Serializer do
...
-    view = Keyword.get(base, parent)
-    acc = Keyword.put(acc, parent, view)
-    handle_include(view, child, acc)
-  end
-  defp handle_include({base, :include}, child, acc) do
-    handle_include(base.relationships(), child, acc)
+  defp get_default_includes(view) do
+    rels = view.relationships()
+    default_includes = rels |> Enum.filter(fn {k, v} ->
+      case v do
```
doomspork: You don't need to use a case here, match on the function head 👍

lib/jsonapi/serializer.ex
```diff
@@ -144,29 +152,31 @@ defmodule JSONAPI.Serializer do
...
+      end
+    end)
   end
-  defp handle_include(base, child, acc) do
-    view = if is_list(base), do: Keyword.get(base, child), else: base
-    Keyword.put(acc, child, view)
+
+  defp get_query_includes(view, query_includes) do
+    rels = view.relationships()
+    Enum.map(query_includes, fn include ->
```
doomspork: I think there is an opportunity to refactor this.  We could probably move to `Enum.reduce/3` and we can eliminate the `case` by matching on the function signature instead.

test/jsonapi_test.exs
```diff
@@ -9,7 +9,7 @@ defmodule JSONAPITest do
     def type, do: "mytype"
     def relationships do
       [author: {JSONAPITest.UserView, :include},
-       other_user: {JSONAPITest.UserView, :include}]
+       other_user: JSONAPITest.UserView]
```
doomspork: Will these changes result in a breaking change?  If not, let's leave the existing tests alone and add new ones to cover these changes.

jlvallelonga: Yeah this changes the functionality. Although, I think it fixes the original intended functionality of `:include` meaning that the relationship will be included by default. I'm basing this on this bit from the readme:

```
  def relationships do
    # The post's author will be included by default
    [author: {MyApp.UserView, :include},
     comments: MyApp.CommentView]
  end
```

Thoughts?

jlvallelonga: @doomspork @jeregrine ping

StareIntoTheBeard:
@doomspork @jeregrine @jlvallelonga bump, any other thoughts on this one? What's needed to get this merged in?

doomspork:
Sorry @jlvallelonga, I've been away on a business trip and didn't have too much time to go through GH.  Since these _are_ a breaking change, I'm going to have to spend some time looking through them closer.  That said, if @jeregrine wants to go forward with it 👍

doomspork:
Thank you @jlvallelonga!  @StareIntoTheBeard I will be cutting a new release shortly 👍

PR #91, opened 2018-01-08
Check headers according to spec close #90

### LINKED ISSUE #90: Improper checking of headers
In [this](https://github.com/jeregrine/jsonapi/blob/1f9e45aee4058ca6b3a8a55aaec6eebcada525a6/lib/jsonapi/plugs/content_type_negotiation.ex) file, it forces the headers to conform to the JSONAPI media type improperly.

To [quote](http://jsonapi.org/format/#content-negotiation-servers):

```
Servers MUST send all JSON API data in response documents with the header Content-Type: application/vnd.api+json without any media type parameters.
```

The server MUST respond with the proper headers, the library handles this okay.

```
Servers MUST respond with a 415 Unsupported Media Type status code if a request specifies the header Content-Type: application/vnd.api+json with any media type parameters.
```

This seems to be worded intentionally, such that **if** a request specifies the `Content-Type: application/vnd.api+json` header **with** any additional parameters, it should return a 415. The server itself is not required to force the client to send the `Content-Type: application/vnd.api+json` header (just to enforce that it's done properly if the client does send it), though [the client is required by the spec to send it](http://jsonapi.org/format/#content-negotiation-clients).

```
Servers MUST respond with a 406 Not Acceptable status code if a request’s Accept header contains the JSON API media type and all instances of that media type are modified with media type parameters.
```

This also seems to be worded intentionally, such that **if** a request specifies an `Accept` header with the proper type, but ALL instances of that media type are modified (meaning it can have multiple media types, and even multiple declarations of the JSONAPI media type), throw a 406. Again, the server is not required to force the client to send the `Accept: application/vnd.api+json` header (just to enforce that it's done properly if the client does send it -- and only to enforce it if they use the `application/vnd.api+json` media type, **and each time it's specified** it has modified media type parameters), and actually, [_the client isn't required to send this header at all, they're just required to **not** modify the media type with parameters on at least one instance of it if they do pass it_](http://jsonapi.org/format/#content-negotiation-clients).

Just wanted to clarify that. Great library!

snewcomer:
I think this is close, but not exactly what we want.  

#90


lib/jsonapi/plugs/content_type_negotiation.ex
```diff
@@ -49,11 +49,18 @@ defmodule JSONAPI.ContentTypeNegotiation do
     {conn, content_type}
   end
 
+  defp respond({conn, @jsonapi, nil}) do
+    before_send(conn)
+  end
   defp respond({conn, @jsonapi, @jsonapi}) do
+    before_send(conn)
+  end
+  defp respond({conn, @jsonapi, _accepts}), do: send_error(conn, 406)
```
snewcomer: I think one part this is missing is IF and ONLY IF "accepts" contains the json api media type, then throw 406.  Otherwise, if sent "application/json", don't throw 406.  Is this correct?

test/jsonapi/plugs/content_type_negotiation_test.exs
```diff
@@ -14,17 +14,39 @@ defmodule JSONAPI.ContentTypeNegotiationTest do
     assert 415 == conn.status
   end
 
-  test "halts and returns an error if accept header is incorrect" do
+  test "halts and returns an error if content-type header contains other media type params" do
+    conn =
+      :post
+      |> conn("/example", "application/vnd.api+json; charset=utf-8")
```
doomspork: This is just setting the post body isn't it?  Wouldn't it be clearer to do something like `Plug.Conn.put_req_header("content-type", "text/html")`?

lib/jsonapi/plugs/content_type_negotiation.ex
```diff
@@ -49,11 +49,18 @@ defmodule JSONAPI.ContentTypeNegotiation do
...
+  defp respond({conn, _content_type, _accepts}), do: send_error(conn, 415)
```
doomspork: I think this need to change.  The content_type isn't required but if it is provided and doesn't match `@jsonapi` we return 415.  Same applies for the accepts header, it can be blank.

lib/jsonapi/plugs/content_type_negotiation.ex
doomspork: Check out this PR https://github.com/emberjs/data/issues/3880

test/jsonapi/plugs/content_type_negotiation_test.exs
```diff
@@ -4,35 +4,101 @@ defmodule JSONAPI.ContentTypeNegotiationTest do
...
+  end
+
+  test "halts and returns an error if content-type header contains other media type params" do
+    conn =
+      :post
+      |> conn("/example", "")
+      |> Plug.Conn.put_req_header("content-type", "application/vnd.api+json; version=1.0")
+      |> Plug.Conn.put_req_header("accept", "application/vnd.api+json")
+      |> ContentTypeNegotiation.call([])
+
```
snewcomer: I think I covered all the test cases here.  Lmk if you see something more! :thought_balloon:

doomspork:
@snewcomer I do believe this covers all the bases!  @TrevorHinesley since you originally reported this, did you want to take a peek at the fix?

Ebtoulson:
Wouldn't the spec also allow wildcards for the accept header?

```
"application/vnd.api+json"
"application/*"
"*/*"
```

TrevorHinesley:
@doomspork sorry about the late reply! And @Ebtoulson that's a great question, I can't tell. I'm not an expert on the spec, but I believe it's language is open to allow *some* leeway like I interpreted in my originally posted Issue. 

If someone else is an expert on the spec, that'd be nice to hear from them. For all I know, even with the way they worded the spec, it IS supposed to require all of this stuff lol. From how the spec seems to read though, I think this is on the right track. 

Also, the `halts and returns an error if all accept header media types contain media type params` test, I think it's worth adding a test that makes sure it SHOULD allow this scenario:

`application/vnd.api+json; version=1.0, application/vnd.api+json;`

But SHOULD NOT allow:

`application/vnd.api+json; version=1.0, application/vnd.api+json; version=1.0`

And also, it SHOULD NOT allow this:

`application/json-or-something-else; application/vnd.api+json; version=1.0, application/vnd.api+json; version=1.0` <-- in this case, not ALL media types have media params, but all of the JSONAPI media types do, which violates spec.

snewcomer:
these are good ones to add to the test cases:

- `application/vnd.api+json; version=1.0, application/vnd.api+json;`

And regarding wildcard - that is a good thought.  I read it as, if sent, it must match `application/vnd.api+json`...but I have been known to not know how to read at time :)

lib/jsonapi/plugs/content_type_negotiation.ex
```diff
@@ -33,8 +33,6 @@ defmodule JSONAPI.ContentTypeNegotiation do
     accepts =
       conn
       |> get_req_header("accept")
-      |> Enum.flat_map(&(String.split(&1, ",")))
-      |> Enum.map(&String.strip/1)
```
snewcomer: removed these lines b/c we want to check to whole string to see if it contains the header we are looking for

lib/jsonapi/plugs/content_type_negotiation.ex
```diff
@@ -49,11 +47,32 @@ defmodule JSONAPI.ContentTypeNegotiation do
...
+    end
+  end
+  defp respond({conn, nil, accepts}) do
+    case validate_header(accepts) do
+      true -> add_header_to_resp(conn)
+      false -> send_error(conn, 406)
+    end
+  end
+  defp respond({conn, @jsonapi, _accepts}), do: send_error(conn, 406)
+  defp respond({conn, _content_type, _accepts}), do: send_error(conn, 415)
```
snewcomer: not sure if this is getting out of hand or just how it is supposed to be :man_shrugging:

snewcomer: I think I need to add another case where both content-type and accept header are passed and need to parse through each one, throwing 415 or 406, whichever one hits first.

snewcomer: which can then replace the last two pattern matched functions

snewcomer:
@doomspork Any idea why travis is failing?

doomspork:
@snewcomer it _might_ be these Credo suggestions:

```shell
  Refactoring opportunities                                                     
┃ 
┃ [F] → Pipe chain should start with a raw value.
┃       lib/jsonapi/serializer.ex:88 #(JSONAPI.Serializer.build_relationships)
┃ [F] → Pipe chain should start with a raw value.
┃       lib/jsonapi/serializer.ex:168 #(JSONAPI.Serializer.get_query_includes)
```

snewcomer:
Ah it would help if I rebased :duck:

doomspork:
@snewcomer I think we're good to merge this, we can always improve it later.  Any thoughts?

snewcomer:
@doomspork I think so.  I pulled it into an app and worked a-ok!

PR #92, opened 2018-01-09
Add to view custom attribute close #89

### LINKED ISSUE #89: Feature Request - custom attribute method
If I define methods in my `**view.ex` file, I want to include them in the attributes hash.  Once could imagine combining `first + last` or some type of conditional logic for a specific field at the view layer.

I was wondering your thoughts were on this and if it is something you want me to take a stab at...

```
defmodule MyApp.ArticleSerializer do

  def fields, do: [:body, :excerpt]

  def excerpt(article, _conn) do
    [first | _ ] = String.split(article.body, ".")
    first
  end
end
```

snewcomer:
Ref #89

### DIFF
--- a/lib/jsonapi/view.ex
+++ b/lib/jsonapi/view.ex
@@ -91,7 +91,19 @@ defmodule JSONAPI.View do
       end
 
       def attributes(data, conn) do
-        visible_fields(data)
+        visible_fields = fields() -- hidden()
+
+        Enum.reduce(visible_fields, %{}, fn field, intermediate_map ->
+          try do
+            merge(intermediate_map, field, apply(__MODULE__, field, [data, conn]))
+          rescue
+            _e -> merge(intermediate_map, field, Map.get(data, field))
+          end
+        end)
+      end
+
+      defp merge(intermediate_map, field, value) do
+        Map.put(intermediate_map, field, value)
       end
 
       def meta(_data, _conn), do: nil
@@ -142,8 +154,6 @@ defmodule JSONAPI.View do
         def render("index.json", %{data: data, conn: conn, params: params}),
           do: index(data, conn, params)
       end
-
-      defp visible_fields(data), do: Map.take(data, fields() -- hidden())
 
       defp host(conn), do: Application.get_env(:jsonapi, :host, conn.host)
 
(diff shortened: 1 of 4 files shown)


lib/jsonapi/view.ex
```diff
@@ -91,7 +91,19 @@ defmodule JSONAPI.View do
...
+        Enum.reduce(visible_fields, %{}, fn field, intermediate_map ->
+          try do
+            merge(intermediate_map, field, apply(__MODULE__, field, [data, conn]))
+          rescue
+            _e -> merge(intermediate_map, field, Map.get(data, field))
+          end
+        end)
+      end
+
+      defp merge(intermediate_map, field, value) do
```
doomspork: What does `merge/3` offer us over just using `Map.put/3` directly?

lib/jsonapi/view.ex
```diff
@@ -91,7 +91,19 @@ defmodule JSONAPI.View do
       end
 
       def attributes(data, conn) do
-        visible_fields(data)
+        visible_fields = fields() -- hidden()
+
+        Enum.reduce(visible_fields, %{}, fn field, intermediate_map ->
```
doomspork: Why `reduce/3` vs `map/2`?

lib/jsonapi/view.ex
```diff
@@ -91,7 +91,19 @@ defmodule JSONAPI.View do
...
+          try do
+            merge(intermediate_map, field, apply(__MODULE__, field, [data, conn]))
```
doomspork: Can we do something like this to DRY things up a bit?

```elixir
value =
  try do
    apply(__MODULE__, field, [data, conn])
  rescue
    _e -> Map.get(data, field)
  end

Map.put(intermediate_map, field, value)
```

lib/jsonapi/view.ex
snewcomer: This was to ensure the return type was a Map, which is what `Map.take` was doing before.

lib/jsonapi/view.ex
snewcomer: ah much better :100:

doomspork:
Thank you @snewcomer!  I want to review #87 and then we can cut a new release 😁

PR #93, opened 2018-01-09
updates plug to allow data with only relationships

jlvallelonga:
Fixes an issue with the `FormatRequired` plug where an error was returned for a valid resource object.

Specifically, a "data" object with only a "relationships" member was not allowed before this change.

see: http://jsonapi.org/format/#document-resource-objects

I found this while trying to patch only relationships (no attribute changes) for a resource like this: http://jsonapi.org/format/#crud-updating-resource-relationships


doomspork:
Thank you @jlvallelonga 👍

Issue #96, opened 2018-02-25
Ability to pass data to meta not apart of payload


snewcomer:
Currently, meta is included in each document payload.  How about including it as a top level document member?  I'd be happy to make a PR.

http://jsonapi.org/format/#document-top-level
http://jsonapi.org/format/#document-meta

snewcomer:
Also, this may be a good time to rethink how meta data is done.  Would would potentially be nice is to pass the meta data as a parameter to `render` and have it included in the payload instead of also passing it through to the view.  What are your thoughts?

PR #97, opened 2018-02-25
include meta as top level document member

snewcomer:
This PR adds the ability to pass a `meta` map to the render function.

The current functionality relies on the meta data to be present on the `data` attribute.  However, for example, on index routes, `data` is simply a List and some attributes to filter/paginate the list may need to be passed through to the response that aren't implicitly included/apart of the `data`.

Lmk your thoughts!

Ref #96

### DIFF
--- a/lib/jsonapi/view.ex
+++ b/lib/jsonapi/view.ex
@@ -62,7 +62,7 @@ defmodule JSONAPI.View do
 
   ## Options
     * `:host` (binary) - Allows the `host` to be overrided for generated URLs.  Defaults to `host` of the supplied `conn`.
-    
+
     * `:scheme` (atom) - Enables configuration of the HTTP scheme for generated URLS.  Defaults to `scheme` from the provided `conn`.
 
     * `:underscore_to_dash` (boolean) - Use dash (`-`) as the word separated for JSON in place of underscore (`_`) per the JSONAPI spec [recommendations](http://jsonapi.org/recommendations/).  Defaults to `false`.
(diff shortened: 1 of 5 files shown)


snewcomer:
Another thought is ensure a user defines a key to put on the `conn` struct and if present, attach to the payload instead of adding another argument to serialize.

lib/jsonapi/view.ex
```diff
@@ -142,15 +142,19 @@ defmodule JSONAPI.View do
       end
 
       if Code.ensure_loaded?(Phoenix) do
+        def render("show.json", %{data: data, conn: conn, params: params, meta: meta}),
+          do: show(data, conn, params, meta: meta)
+        def render("show.json", %{data: data, conn: conn, meta: meta}),
```
snewcomer: One unrelated curiosity: 

If I define a render function in my controller, how does the `conn` parameter get in there?

```
render(conn, "index.json", %{data: my_list, meta: meta})
```

jeregrine: conn is always merged into the assigns data, because the view doesn't mutate/return a conn its treated as data instead of the subject of the function.

snewcomer:
@jeregrine Looks like 18.3 is failing here and on master.  Only on elixir 1.6 is otp 18 not supported.  Would you like me to remove `18.3` from travis?  Or is there another way to do this?

https://github.com/elixir-lang/elixir/blob/master/lib/elixir/pages/Compatibility%20and%20Deprecations.md#compatibility-between-elixir-and-erlangotp

https://travis-ci.org/jeregrine/jsonapi/jobs/347070137
https://travis-ci.org/jeregrine/jsonapi/jobs/346951834

jeregrine:
Yea we just need to update the travis matrix a good example is
https://github.com/getsentry/sentry-elixir/blob/master/.travis.yml

doomspork:
@snewcomer could you squash your commits to re-trigger Travis? 😁

doomspork:
@jeregrine do you want me to cut a new release?

jeregrine:
I trust your judgement here :)

PR #99, opened 2018-02-27
Updates and new version

doomspork:
@jeregrine this removes some warnings that exist, applies the formatter, updates TravisCI, and sets the version to `0.7.0`


doomspork:
@jeregrine this is released 👍

PR #101, opened 2018-03-06
Fix show meta

snewcomer:
This is my fault.  I was trying to figure out for the longest time why the `show` meta was not working properly.  Looking for a test to write to back this up.

Improves on https://github.com/jeregrine/jsonapi/pull/97

### DIFF
--- a/lib/jsonapi/view.ex
+++ b/lib/jsonapi/view.ex
@@ -180,22 +180,27 @@ defmodule JSONAPI.View do
         "#{url_for(data, conn)}?#{query}"
       end
 
-      if Code.ensure_loaded?(Phoenix) do
-        def render("show.json", %{data: data, conn: conn, params: params, meta: meta}),
-          do: show(data, conn, params, meta)
+      with true <- Code.ensure_loaded?(Phoenix)
+        do
+          def render("show.json", %{data: data, conn: conn, params: params, meta: meta}),
+            do: show(data, conn, params, meta)
 
-        def render("show.json", %{data: data, conn: conn, meta: meta}),
-          do: show(data, conn, conn.params, meta)
+          def render("show.json", %{data: data, conn: conn, meta: meta}),
+            do: show(data, conn, conn.params, meta)
 
-        def render("show.json", %{data: data, conn: conn}), do: show(data, conn, conn.params)
+          def render("show.json", %{data: data, conn: conn}), do: show(data, conn, conn.params)
 
-        def render("index.json", %{data: data, conn: conn, params: params, meta: meta}),
-          do: index(data, conn, params, meta)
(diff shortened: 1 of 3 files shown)


snewcomer:
It would be great to add some tests around this [behavior](https://github.com/jeregrine/jsonapi/blob/master/lib/jsonapi/view.ex#L152).  Not seeing anything yet.

snewcomer:
Added some tests.  Although a bit mystified on the failures on 1.6.  It seems something to do with the formatter, but I may be missing some knowledge of what to do.

jeregrine:
I believe you just need to run the formatter on the test file.

snewcomer:
Ah thank you.  I wasn't on 1.6 previously!  So one other problem is the console output.  `Redefining module JSONAPI.ViewTest.CommentView (current version defined in memory)` && `redefining module Phoenix...`

I'm wondering how we can load Phoenix once for the CommentView in the test.  Any thoughts?  Maybe a test helper macro?

https://travis-ci.org/jeregrine/jsonapi/jobs/349849195

jeregrine:
I wonder if setup tags is even required.

snewcomer:
@jeregrine That is a good idea.  So if I remove `setup tags`, I could also remove [this](https://github.com/jeregrine/jsonapi/blob/master/lib/jsonapi/view.ex#L143) and all tests pass; however, I am also lacking context on why that is needed and what various use cases `ensure_loaded` covers.

snewcomer:
@jeregrine So I added Phoenix as a test dep.  Lmk what you think.  Also tested the render functions as well.

lib/jsonapi/view.ex
```diff
@@ -140,12 +140,12 @@ defmodule JSONAPI.View do
         "#{url_for(data, conn)}/relationships/#{rel_type}"
       end
 
-      if Code.ensure_loaded?(Phoenix) do
+      with true <- Code.ensure_loaded?(Phoenix) do
```
doomspork: Why `with` over an `if` here?  I'm not sure we gain anything.

snewcomer: @doomspork Added `raise ArgumentError` in the `else` case.  Should we keep that or remove it?  Don't mind either way, but thought, in a very rare case, might help somebody.

doomspork:
@snewcomer any updates here?

snewcomer:
@jeregrine @doomspork This PR is ready for your re-review!

doomspork:
@jeregrine I do want to merge with your review still showing rejected.  Do you want to dismiss or re-review?

doomspork:
Thank you @jeregrine ❤️

jeregrine:
Thank you all! I will try to cut release tonight.

PR #102, opened 2018-04-15
Apply optional dash-to-underscore to include keys

mvrkljan:
When specifying an include string in request which contains dashed entries (e.g. `article.guest-authors`) and library is configured to convert dashes to underscore `guest-authors` should be transformed to `guest_authors`. 

Currently, without this an error is thrown caused by converting string to an unknown atom.


lib/jsonapi/utils/dash.ex
```diff
@@ -0,0 +1,29 @@
+defmodule JSONAPI.Utils.Dash do
```
jeregrine: Does this do more than the existing https://github.com/jeregrine/jsonapi/blob/d4371362ecaadaece27f4d640722f0beae90e6ad/lib/jsonapi/utils/underscore.ex#L1 ?

mvrkljan: :ping_pong: @jeregrine

jeregrine: Lets merge this into one module, if you feel better renaming it thats fine just make sure the tests pass. But since they both use the same configuration and both have functionally the same responsibility I think having one will be better.

mvrkljan:
@jeregrine updated, thanks!

mvrkljan:
@jeregrine Anything else needed for this one?

jeregrine:
Thank you for the work on this! Sorry for the delay ❤️

PR #103, opened 2018-04-15
Do not build relationships section for not loaded relationships

mvrkljan:
Currently if a resource has defined relationships on its view it is always generated an record in the `relationships` section. If the relationship is not loaded, it generates a relationship record with `"id": null` and record type. 

This is not precisely per spec and can cause weird issues with front-end libraries when a record appears in relationships section with an id value of null.

Also, in case of one-to-many relationships, having a single record with id null can be misleading.

This PR updates the logic to not include an empty record if the relationship is not loaded.


jeregrine:
Thanks! 💟

PR #106, opened 2018-05-01
Dasherize keys in relationship urls

wasnotrice:
Hi, thanks for this library, it's really nice to use!

I noticed that the keys that become part of relationship urls weren't getting dashed. So this is a small test and fix. Should be compatible with #102.

Cheers!

### DIFF
--- a/lib/jsonapi/serializer.ex
+++ b/lib/jsonapi/serializer.ex
@@ -81,12 +81,13 @@ defmodule JSONAPI.Serializer do
 
     only_rel_view = get_view(rel_view)
     # Build the relationship url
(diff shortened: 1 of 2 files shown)


jeregrine:
Thank you for this but I believe its a duplicate of an un-released pr https://github.com/jeregrine/jsonapi/pull/102

wasnotrice:
@jeregrine It's related, but it's not a duplicate. On master, after #102 was merged, if you run the test added here, the relationship's self url is

```
"/mytype/1/relationships/best_comments"
```

when it should be

```
"/mytype/1/relationships/best-comments"
```

jeregrine:
Thanks! ❤️

PR #111, opened 2018-05-29
fixing example in readme

gushonorato:


gushonorato:
@jeregrine I'm using Phoenix 1.3.2 and I can't get JSON API working without including this two methods in my view. Am I doing something wrong? Here is my view:

```
defmodule DrivemeterApiWeb.TransactionView do
  use JSONAPI.View, type: "transactions"
  alias DrivemeterApiWeb.TransactionView

  def fields do
    [:amount,:category,:description,:distance,:duration,:incurred_at]
  end

  def render("index.json", %{transactions: transactions, conn: conn}) do
    TransactionView.index(transactions, conn, conn.params)
  end
  
  def render("show.json", %{transaction: transaction, conn: conn}) do
    TransactionView.show(transaction, conn, conn.params)
  end
end
```

When I remove these two render methods above, I'm getting this error: `no function clause matching in DrivemeterApiWeb.TransactionView.render/2`

jeregrine:
@gushonorato you appear to have found a bug. If possible could you try and reproduce this in a test and or fix the bug? If not I will take a look.

doomspork:
@jeregrine / @gushonorato I'm not 💯 this is a bug.  If you look at the code we generate it, it's looking for `data:` and you're relying on `transaction`.  Could you try changing that and see if you still need those methods?

PR #114, opened 2018-06-04
Handle dashed include for top-level relationship

wasnotrice:
Dashed includes were supported for nested relationships, but not for top-level ones. Also removes a stray `String.to_atom`


doomspork:
💥 Thank you @wasnotrice

PR #121, opened 2018-06-30
Xarango id is in _key not id

jnylen:
This will grab the _key (which is id in arangodb) and use it as the id.

_id in arangodb is collection/key

### DIFF
--- a/lib/jsonapi/view.ex
+++ b/lib/jsonapi/view.ex
@@ -98,6 +98,7 @@ defmodule JSONAPI.View do
 
       def id(nil), do: nil
       def id(%{__struct__: Ecto.Association.NotLoaded}), do: nil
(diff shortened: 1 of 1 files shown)


doomspork:
@jnylen thanks for the PR but I don't think we want to include this change in the core library.   

If we try to handle all possible cases for `id/1` we'll forever be adding new function signatures.  To address this, we allow you to override the `id/1` function to handle things specific to your application like this! 
 https://github.com/jnylen/jsonapi/blob/e161ef927abd53d2f116229dc1a8e98c40951a58/lib/jsonapi/view.ex#L202

Happy to answer any questions you have about implementing this in your project 👍

jnylen:
@doomspork,
Oh, I totally missed that variant. I'll try that instead. Thanks for a great library!

PR #124, opened 2018-08-03
Omit non-Object meta

kbaird:
The serializer was including `meta` even in cases where its value would not be an Object, which is not in compliance with the JSONAPI spec (http://jsonapi.org/format/#document-meta). This PR makes the serializer omit the `meta` when `is_map(meta)` is `false`.

I also DRYed up the tests, since my added ones re-used some of the existing default data.

### DIFF
--- a/lib/jsonapi/query_parser.ex
+++ b/lib/jsonapi/query_parser.ex
@@ -117,8 +117,7 @@ defmodule JSONAPI.QueryParser do
       requested_fields =
         value
         |> String.split(",")
-        |> Enum.map(&String.to_atom/1)
-        |> Enum.into(MapSet.new())
+        |> Enum.into(MapSet.new(), &String.to_atom/1)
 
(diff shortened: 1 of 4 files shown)


mjc:
I think this is a bugfix rather than an enhancement as `"meta": null` was the previous output and that is not allowed by the spec.

doomspork:
Thank you and welcome @kbaird 😁 

I've corrected the label

lib/jsonapi/query_parser.ex
```diff
@@ -117,8 +117,7 @@ defmodule JSONAPI.QueryParser do
       requested_fields =
         value
         |> String.split(",")
-        |> Enum.map(&String.to_atom/1)
-        |> Enum.into(MapSet.new())
+        |> Enum.into(MapSet.new(), &String.to_atom/1)
```
snewcomer: 👍

PR #127, opened 2018-08-30
refactor: Content-type separtor use ';' instead of ','

toyah:
ref: https://www.w3.org/Protocols/rfc1341/4_Content-Type.html

### DIFF
--- a/lib/jsonapi/plugs/content_type_negotiation.ex
+++ b/lib/jsonapi/plugs/content_type_negotiation.ex
@@ -62,7 +62,7 @@ defmodule JSONAPI.ContentTypeNegotiation do
   end
 
   defp validate_header(string) when is_binary(string) do
-    string |> String.split(",") |> Enum.map(&String.trim/1) |> Enum.member?(@jsonapi)
+    string |> String.split(";") |> Enum.map(&String.trim/1) |> Enum.member?(@jsonapi)
   end
 
   defp validate_header(nil), do: true


doomspork:
Thank you @toyah!

snewcomer:
@doomspork @jeregrine I have a feeling this PR might need to be reverted.  We separate multiple content-type or accept headers with a `,` and then use the `;` to separate the `type` and `parameter`.  As a result, the content_type_negotiation tests are failing on master.  I believe we already follow the spec:

From [Content-Type](https://www.w3.org/Protocols/rfc1341/4_Content-Type.html):
`type "/" subtype *[";" parameter]`

So we allow:
`application/vnd.api+json; version=1.0, application/vnd.api+json`

And then:

http://jsonapi.org/format/upcoming/#content-negotiation-clients

> Note: The content negotiation requirements exist to allow future versions of this specification to use media type parameters for extension negotiation and versioning.

doomspork:
@snewcomer is it the change that's wrong or our content negotiation?  I'm pretty tied up this week so I won't have much time to dive into the spec but I'm confident in the decision you make so happy to defer to your judgement.

Thoughts @jeregrine?

jeregrine:
I am happy to defer to @snewcomer's judgment

snewcomer:
@toyah The link you put in the description is correct and we do use `;` to separate the type/subtype && the parameter.  The specific code this PR touched (the `,`) was to separate multiple media types, which the spec does provide leeway on.  So:

__Valid Content-Type/Accept header__

`application/vnd.api+json; version=1.0`
`application/vnd.api+json; version=1.0, application/vnd.api+json`

and we split on `,` to ensure that at least one of the media types has `application/vnd.api+json` in it.

Were you running into a specific error or perhaps do you see something I am not?

PR #128, opened 2018-09-05
Revert "refactor: Content-type separtor use ';' instead of ',' (#127)"

### LINKED ISSUE #123: fix: Content-type separtor use ';' instead of ','
https://www.w3.org/Protocols/rfc1341/4_Content-Type.html

snewcomer:
This reverts commit 2694b02b2a87bbbebcb6045a65d2bafa30fc0f64.

@toyah Lmk if you have any thoughts

See #127 for discussion on current use of `;` and `,`

Close #123


doomspork:
@toyah I too would like to hear your thoughts before we merge this 👍

snewcomer:
I think we can probably go ahead and merge this if that is ok with ya'll!

PR #129, opened 2018-09-12
Removes warning about hidden being undefined

zamith:
Why:

* When compiling a view that used this module it would throw a warning
  saying that `hidden/0` was undefined

* Tricking the compiler to now throw the warning since we know that it
  will be defined as we're only calling it conditionally on the fact that
  it has been overridden.


zamith:
@doomspork Rebased with master, which should fix the errors. Can we merge it?

PR #130, opened 2018-09-12
Globally configurable namespace

zamith:
Why:

* A lot of the times a namespace is defined is to add the `/api` prefix
  to all the routes, and having to add it to all the views becomes tedious
  and error prone

* Allowing the namespace to be defined in a global config as a default
  that can be changed on a module by module basis

### DIFF
--- a/lib/jsonapi/view.ex
+++ b/lib/jsonapi/view.ex
@@ -88,7 +88,8 @@ defmodule JSONAPI.View do
   """
   defmacro __using__(opts \\ []) do
     {type, opts} = Keyword.pop(opts, :type)
-    {namespace, _opts} = Keyword.pop(opts, :namespace, "")
+    {module_namespace, _opts} = Keyword.pop(opts, :namespace)
+    namespace = module_namespace || Application.get_env(:jsonapi, :namespace, "")
 
     quote do
       import JSONAPI.Serializer, only: [serialize: 4]
--- a/test/jsonapi/view_test.exs
+++ b/test/jsonapi/view_test.exs
@@ -1,8 +1,6 @@
 defmodule JSONAPI.ViewTest do
   use ExUnit.Case
 
-  Application.put_env(:jsonapi, :namespace, "/other-api")
-
   setup tags do
     if tags[:compile_phoenix] do
       Module.create(Phoenix, [], __ENV__)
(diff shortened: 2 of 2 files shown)


doomspork:
@zamith can you take a peek at the build?  It looks like tests are failing.

zamith:
@doomspork The issue with testing this is that I'm doing the namespace check at compile time, which makes it impossible to change for one test or even test file only. The option here would be to add a module namespace option to all the views or have different view names in the serializer and view tests.

Another option would be to move the check to runtime, but that's an unnecessary call each time we generate a URL.

What do you think?

doomspork:
My sincerest apologizes @zamith!  I got caught up with family things and this PR slipped through the cracks.

I'm not following.  I'm just asking that you investigate why the Travis CI build is failing on this PR before we merge.  Are you suggesting these changes are going to cause our tests to fail?

lib/jsonapi/view.ex
```diff
@@ -88,7 +88,8 @@ defmodule JSONAPI.View do
   """
   defmacro __using__(opts \\ []) do
     {type, opts} = Keyword.pop(opts, :type)
-    {namespace, _opts} = Keyword.pop(opts, :namespace, "")
+    {module_namespace, _opts} = Keyword.pop(opts, :namespace)
+    namespace = module_namespace || Application.get_env(:jsonapi, :namespace, "")
```
jherdman: A humble request: would you mind documenting this change in the README (or elsewhere) too?

doomspork:
@zamith would you mind adding a note to the README about this change? It would be a wonderful addition to include in a new release along with the other recently added features.

jherdman:
@zamith I figured I'd check in with you again. If you find yourself a bit swamped I could pick up where you left off (if you'd like).

jherdman:
@doomspork @jeregrine I propose that, given the value of this feature, we take over this PR. I can submit a working version this afternoon if we're in agreement.

@zamith please let me know if you'd like to take this over the finish line otherwise.

doomspork:
@jherdman sounds good to me. You should be able to work off of @zamith's commits so he can still get some credit 😁

jherdman:
Closing in favour of #178.

PR #135, opened 2018-10-02
Remove leading is_ from is_data_loaded?

kbaird:
Style preferences are
- _is_predicate_ for guard-safe macros and
- _predicate?_ for regular boolean functions.


doomspork:
Thank you @kbaird!  @jeregrine any objections?  I'm all for this.  We'll need to bump the version of the new release you cut to account for the breaking change.

kbaird:
You're welcome, @doomspork. I have a separate question about custom relationships in a View. Where's the best venue for that?

Issue #132, opened 2018-09-14
Add option to exclude keys from underscore/dash treatment?


wasnotrice:
I'm working with an API where I have a field `data_schema` that is an arbitrary JSON object. The API uses the dashed key format, so it shows up in requests/responses as `data-schema`. This is all fine until someone includes an underscore in one of their JSON object keys, such as `__finalized__`. So my incoming data

```elixir
%{
  "data" => %{
    "attributes" => %{
      "data-schema" => %{
        "properties" => %{
          "__finalized__" => %{"title" => "Finalized", "type" => "boolean"},
        },
        "type" => "object"
      },
    },
    "relationships" => %{},
    "type" => "schemas"
  }
}
```

gets transformed into

```elixir
%{
  "data" => %{
    "attributes" => %{
      "data-schema" => %{
        "properties" => %{
          "--finalized--" => %{"title" => "Finalized", "type" => "boolean"}
        },
        "type" => "object"
      },
      "inserted-at" => "2018-09-14T21:03:25.829318",
      "updated-at" => "2018-09-14T21:03:25.829328"
    },
    "id" => "27304",
    "relationships" => %{},
    "type" => "schemas"
  },
  "included" => [],
  "meta" => nil
}
```

(this is surprising to the people who used the leading underscores to indicate "don't mess with this") Ideally, the value of `data-schema` would not be touched at all. I'm not sure how we could accomplish that simply. But I wonder if it makes sense to add an underscore option that specifies field prefixes or patterns that would not be underscored/dashed? Any thoughts?

doomspork:
@wasnotrice does `underscore_to_dash: false` not do the trick here?  After a quick glance through the code I believe the only time we change anything is if this is set to `true`

wasnotrice:
@doomspork sorry I don't think I'm doing a very good job of explaining :)

The api is generated with this library, so the reason that the keys are dashed in the first place is that I have `underscore_to_dash: true`. That is desired.

But given that my api serves responses with dashed keys, when some of my fields are arbitrary JSON, I don't want the library to descend into that arbitrary JSON and dash the keys there, too. I want to say something like "transform keys to dashes in this map, but skip the value of `data_schema`"

doomspork:
Ah ha!  I follow now.  I'm not 💯on how best to accomplish this though 🤔

Maybe we could support 4 options for `underscore_to_dash`: `true`, `false`, `only: []` and `except: []`, with the latter two implying `true`.

Something like:

```elixir
[...]

PR #137, opened 2018-10-16
WIP: Allow :only and :except when enabling underscore_to_dash

wasnotrice:
Fix for #132 

This adds the option to specify

```elixir
underscore_to_dash:
  only: [:all_my, :special_keys]
```

or

```elixir
underscore_to_dash:
  except: [:all_my, :special_keys]
```

For serialization the tests are pretty solid. I haven't built out the various edge case tests for query parsing yet. Thought I'd get some feedback first.

Notes:

1. I modified the `underscore` functions to accept a view, but the view is just ignored right now. I was thinking it would be nice to be able to specify keys per-view. I couldn't come up with a great way to configure that, so that view just gets passed along, and doesn't do anything yet. But it's ready and waiting. One option would be to allow for `{ViewModule, :special_key}` tuples in the Application.env lists. Any thought on this?

2. The current behavior is that when the config specifies keys, both the key and its value gets the same treatment. So if a key is in the `only` list, then it and its value are run through the `underscore` function. If not, neither the key nor the value are. Similarly for the `except` list.

3. I haven't done any testing for the scenario where a user might specify both `only` and `except`, but if `only` is present, it will win, and except will be ignored.

4. I added some `describe` blocks around the tests that modify `Application.env`, so that we can unset the changes in an on_exit block. This makes the tests more reliable when there are failures (i.e. the Application.env doesn't leak into other tests)

### DIFF
--- a/lib/jsonapi/utils/underscore.ex
+++ b/lib/jsonapi/utils/underscore.ex
@@ -3,7 +3,7 @@ defmodule JSONAPI.Utils.Underscore do
   Helpers for replacing underscores with dashes.
   """
 
-  def underscore?, do: Application.get_env(:jsonapi, :underscore_to_dash, false)
+  def underscore?, do: Application.get_env(:jsonapi, :underscore_to_dash, false) != false
 
   @doc """
   Replace dashes between words in `value` with underscores
@@ -42,9 +42,13 @@ defmodule JSONAPI.Utils.Underscore do
       "_top__posts_"
   """
   def underscore(value) when is_atom(value) do
-    value
-    |> to_string
-    |> underscore
+    if underscore?(value) do
+      value
+      |> to_string
+      |> underscore()
+    else
+      to_string(value)
+    end
   end
 
   def underscore(value) when is_binary(value) do
@@ -56,14 +60,44 @@ defmodule JSONAPI.Utils.Underscore do
   end
 
   def underscore(value) when is_map(value) do
-    Enum.into(value, %{}, &underscore/1)
+    Enum.into(value, %{}, &underscore(&1))
   end
 
   def underscore({key, value}) do
-    if is_map(value) do
-      {underscore(key), underscore(value)}
+    if underscore?(key) do
+      if is_map(value) do
+        {underscore(key), underscore(value)}
+      else
+        {underscore(key), value}
+      end
     else
       {underscore(key), value}
+    end
+  end
+
+  defp underscore?(key) do
+    config = Application.get_env(:jsonapi, :underscore_to_dash)
+    config_specifies_underscore?(config, key)
+  end
+
+  defp config_specifies_underscore?(true, _), do: true
+  defp config_specifies_underscore?(false, _), do: false
+
+  defp config_specifies_underscore?(config, key) when is_list(config) do
+    if Keyword.has_key?(config, :only) do
+      config_specifies?(config, :only, key)
+    else
+      !config_specifies?(config, :except, key)
+    end
+  end
+
+  defp config_specifies?(config, only_or_except, key) when is_list(config) do
+    case Keyword.get(config, only_or_except) do
+      list when is_list(list) ->
+        Enum.member?(list, key)
+
+      _ ->
+        false
     end
   end
 end
--- a/lib/jsonapi/utils/underscore.ex
+++ b/lib/jsonapi/utils/underscore.ex
@@ -3,7 +3,7 @@ defmodule JSONAPI.Utils.Underscore do
   Helpers for replacing underscores with dashes.
   """
 
-  def underscore?, do: Application.get_env(:jsonapi, :underscore_to_dash, false) != false
+  def underscore?, do: config() != false
 
   @doc """
   Replace dashes between words in `value` with underscores
@@ -72,8 +72,7 @@ defmodule JSONAPI.Utils.Underscore do
   end
 
   defp underscore?(key) do
-    config = Application.get_env(:jsonapi, :underscore_to_dash)
-    config_specifies_underscore?(config, key)
+    config_specifies_underscore?(config(), key)
   end
 
   defp config_specifies_underscore?(true, _), do: true
@@ -87,4 +86,5 @@ defmodule JSONAPI.Utils.Underscore do
     end
   end
 
+  defp config, do: Application.get_env(:jsonapi, :underscore_to_dash, false)
 end
--- a/lib/jsonapi/utils/underscore.ex
+++ b/lib/jsonapi/utils/underscore.ex
@@ -44,7 +44,7 @@ defmodule JSONAPI.Utils.Underscore do
   def underscore(value) when is_atom(value) do
     if underscore?(value) do
       value
-      |> to_string
+      |> to_string()
       |> underscore()
     else
       to_string(value)
@@ -60,16 +60,12 @@ defmodule JSONAPI.Utils.Underscore do
   end
 
   def underscore(value) when is_map(value) do
-    Enum.into(value, %{}, &underscore(&1))
+    Enum.into(value, %{}, &underscore/1)
   end
 
   def underscore({key, value}) do
-    if underscore?(key) do
-      if is_map(value) do
-        {underscore(key), underscore(value)}
-      else
-        {underscore(key), value}
-      end
+    if underscore?(key) and is_map(value) do
+      {underscore(key), underscore(value)}
     else
       {underscore(key), value}
     end
@@ -91,13 +87,7 @@ defmodule JSONAPI.Utils.Underscore do
     end
   end
 
-  defp config_specifies?(config, only_or_except, key) when is_list(config) do
-    case Keyword.get(config, only_or_except) do
-      list when is_list(list) ->
-        Enum.member?(list, key)
-
-      _ ->
-        false
-    end
+  defp config_specifies?(config, only_or_except, key) do
+    key in Keyword.get(config, only_or_except, [])
   end
 end
--- a/lib/jsonapi/utils/underscore.ex
+++ b/lib/jsonapi/utils/underscore.ex
@@ -80,14 +80,11 @@ defmodule JSONAPI.Utils.Underscore do
   defp config_specifies_underscore?(false, _), do: false
 
   defp config_specifies_underscore?(config, key) when is_list(config) do
-    if Keyword.has_key?(config, :only) do
-      config_specifies?(config, :only, key)
-    else
-      !config_specifies?(config, :except, key)
+    cond do
+      Keyword.has_key?(config, :only) -> key in Keyword.get(config, :only, [])
+      Keyword.has_key?(config, :except) -> key not in Keyword.get(config, :except, [])
+      true -> false
     end
   end
 
-  defp config_specifies?(config, only_or_except, key) do
-    key in Keyword.get(config, only_or_except, [])
-  end
 end
--- a/lib/jsonapi/utils/underscore.ex
+++ b/lib/jsonapi/utils/underscore.ex
@@ -81,7 +81,7 @@ defmodule JSONAPI.Utils.Underscore do
   defp config_specifies_underscore?(config, key) when is_list(config) do
     cond do
       Keyword.has_key?(config, :only) -> key in Keyword.get(config, :only, [])
-      Keyword.has_key?(config, :except) -> key not in Keyword.get(config, :except, [])
+      Keyword.has_key?(config, :except) -> not key in Keyword.get(config, :except, [])
       true -> false
     end
   end
--- a/test/serializer_test.exs
+++ b/test/serializer_test.exs
@@ -41,6 +41,7 @@ defmodule JSONAPISerializerTest do
       ]
 
     def meta(data, _conn), do: %{meta_text: "meta_#{data[:text]}"}
+
     def type, do: "mytype"
 
     def relationships do
--- a/test/serializer_test.exs
+++ b/test/serializer_test.exs
@@ -456,7 +456,9 @@ defmodule JSONAPISerializerTest do
 
   describe "when underscore_to_dash: except" do
     setup do
-      Application.put_env(:jsonapi, :underscore_to_dash,
+      Application.put_env(
+        :jsonapi,
+        :underscore_to_dash,
         except: [:full_description, :_private_attribute, :metadata]
       )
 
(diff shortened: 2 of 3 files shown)


wasnotrice:
@snewcomer thanks for the review. I could definitely break out the regex modification (that's really just changing the regex). I'll do that and then rebase this PR once it's merged.

I realize there are a few different things happening in this PR—I tried to highlight them in the initial comment. Maybe it would be easier to review if I also broke out the improvements to test apparatus for existing tests?

Plus I see that the formatter failed. Boo :)

wasnotrice:
Not sure why the formatter is failing in CI—it passes on 1.7.3

lib/jsonapi/utils/underscore.ex
```diff
@@ -56,14 +60,44 @@ defmodule JSONAPI.Utils.Underscore do
   end
 
   def underscore(value) when is_map(value) do
-    Enum.into(value, %{}, &underscore/1)
+    Enum.into(value, %{}, &underscore(&1))
```
doomspork: For my own edification, what is the purpose of this change?

lib/jsonapi/utils/underscore.ex
```diff
@@ -56,14 +60,44 @@ defmodule JSONAPI.Utils.Underscore do
...
   end
 
   def underscore({key, value}) do
-    if is_map(value) do
-      {underscore(key), underscore(value)}
+    if underscore?(key) do
```
doomspork: Could this be simplified in such a way to remove the need for a nested `if`?

```elixir
if underscore?(key) && is_map(value) do
  {underscore(key), underscore(value)}
else
  {underscore(key), value}
end	   
```

lib/jsonapi/utils/underscore.ex
```diff
@@ -56,14 +60,44 @@ defmodule JSONAPI.Utils.Underscore do
...
+
+  defp config_specifies_underscore?(config, key) when is_list(config) do
+    if Keyword.has_key?(config, :only) do
+      config_specifies?(config, :only, key)
+    else
+      !config_specifies?(config, :except, key)
+    end
+  end
+
+  defp config_specifies?(config, only_or_except, key) when is_list(config) do
```
doomspork: Do we need this this `is_list/1` guard?  There's no other `config_specifies?/3` functions that could be called.  If we intend to handle non-list cases, we should define that function.

I'm not sure about the `case` and `is_list/1` inside either 🤔.  If a user supplies an invalid configuration (a non-list) such that we need `is_list/1` _should_ we proceed quietly?  That seems like it would not be expected by the user.

I'm wondering if we should just simplify this and in the event they pass invalid configuration it'll raise an error:

```elixir
defp config_specifies?(config, only_or_except, key) do
  key in Keyword.get(config, only_or_except, [])
end
```

lib/jsonapi/utils/underscore.ex
```diff
@@ -56,14 +60,44 @@ defmodule JSONAPI.Utils.Underscore do
...
+  defp underscore?(key) do
+    config = Application.get_env(:jsonapi, :underscore_to_dash)
+    config_specifies_underscore?(config, key)
+  end
+
+  defp config_specifies_underscore?(true, _), do: true
+  defp config_specifies_underscore?(false, _), do: false
+
+  defp config_specifies_underscore?(config, key) when is_list(config) do
+    if Keyword.has_key?(config, :only) do
```
doomspork: Would this suffice?

```elixir
  defp config_specifies_underscore?(config, key) do
    config_specifies?(config, :only, key) or not config_specifies?(config, :except, key)
  end
```

Since you're guarding in `config_specifies?/3` and handling an error case, we don't need to check if the either `:only` or `:except` exists we can just call the functions.

test/serializer_test.exs
```diff
@@ -25,6 +25,38 @@ defmodule JSONAPISerializerTest do
...
+        :text,
+        :body,
+        :full_description,
+        :inserted_at,
+        :metadata,
+        :_private_attribute,
+        :nonstandard__attribute
+      ]
+
+    def meta(data, _conn), do: %{meta_text: "meta_#{data[:text]}"}
```
doomspork: Please add a space between these two functions.  Only the same functions (name/arity) should be grouped without newlines.

test/serializer_test.exs
```diff
@@ -410,6 +453,133 @@ defmodule JSONAPISerializerTest do
...
+
+      assert attributes["inserted-at"] == data[:inserted_at]
+      assert attributes["nonstandard__attribute"] == data[:nonstandard__attribute]
+
+      # Not modified
+      assert attributes["full_description"] == data[:full_description]
+      assert attributes["metadata"] == data[:metadata]
+      assert attributes["_private_attribute"] == data[:_private_attribute]
+
+      assert Enum.find(included, fn i -> i[:type] == "user" && i[:id] == "2" end)[:attributes][
```
doomspork: I personally find these `assert` statements very difficult to grok, can we clean them up in any way?  It may be possible to use a pipeline:

```elixir
assert "bonds" == included
                  |> Enum.find(&(&1.type == "user" && &1.id == "2")
                  |> get_in([:attributes, "last-name"])

# Or perhaps with a capture 
last_name = included
            |> Enum.find(&(&1.type == "user" && &1.id == "2")
            |> get_in([:attributes, "last-name"])

assert "bonds" == last_name


```

lib/jsonapi/utils/underscore.ex
```diff
@@ -42,9 +42,13 @@ defmodule JSONAPI.Utils.Underscore do
       "_top__posts_"
   """
   def underscore(value) when is_atom(value) do
-    value
-    |> to_string
-    |> underscore
+    if underscore?(value) do
+      value
+      |> to_string
```
doomspork: `to_string()` please 😁

lib/jsonapi/utils/underscore.ex
wasnotrice: Nice catch—this is leftover from a previous version that added a parameter to the call. Now unnecessary

lib/jsonapi/utils/underscore.ex
wasnotrice: yes!

lib/jsonapi/utils/underscore.ex
wasnotrice: yes!

lib/jsonapi/utils/underscore.ex
wasnotrice: No, in fact that is how I originally wanted to write it, but it is not the same logic. If `:only` is specified, then we do not want to check `:except` at all, regardless of whether the `:only` specifies a key. Especially since `:except` is negated, which means that if neither `:only` nor `:except` is specified, then `not config_specifies?(config, :except, key)` will return true.

test/serializer_test.exs
wasnotrice: yep

test/serializer_test.exs
wasnotrice: I agree! These are modeled after other tests in this file (see [line 441](https://github.com/jeregrine/jsonapi/pull/137/files#diff-e5e7be16724a4eee8799ab05f1d9cc83R441)). I thought it would be better to keep same style while adding tests, to focus on the content of this addition.

Another option would be to add some helper functions to the tests:

```elixir
assert "bonds" ==
  included
  |> find_user_with_id("2")
  |> get_attribute("last-name")
```

Do you want to address this here, or would it be better to do a separate PR to improve readability of these tests?

lib/jsonapi/utils/underscore.ex
wasnotrice: of course :)

lib/jsonapi/utils/underscore.ex
wasnotrice: Here are a couple of alternatives that keep the same logic, and remove  the indirection of `config_specifies/3`:

```elixir
# All in one if
defp config_specifies_underscore?(config, key) when is_list(config) do
  if Keyword.has_key?(config, :only) do
    key in Keyword.get(config, :only, [])
  else
    key not in Keyword.get(config, :except, [])
  end
end
```

```elixir
# Cond
defp config_specifies_underscore?(config, key) when is_list(config) do
  cond do
    Keyword.has_key?(config, :only) -> key in Keyword.get(config, :only, [])
    Keyword.has_key?(config, :except) -> key not in Keyword.get(config, :except, [])
    true -> false
  end
end
```

Is either of those alternatives preferable?

snewcomer: second one looks good!

test/serializer_test.exs
snewcomer: Agree with the pipeline style for these tests.  And I would imagine another PR would be preferable.

lib/jsonapi/utils/underscore.ex
```diff
@@ -3,7 +3,7 @@ defmodule JSONAPI.Utils.Underscore do
   Helpers for replacing underscores with dashes.
   """
 
-  def underscore?, do: Application.get_env(:jsonapi, :underscore_to_dash, false)
+  def underscore?, do: Application.get_env(:jsonapi, :underscore_to_dash, false) != false
```
snewcomer: Should we make the config a module attribute above and use it here and down in the `underscore?/1`?

wasnotrice: Sure—good idea

test/serializer_test.exs
```diff
@@ -25,6 +25,39 @@ defmodule JSONAPISerializerTest do
...
+    use JSONAPI.View
+
+    def fields,
+      do: [
+        :text,
+        :body,
+        :full_description,
+        :inserted_at,
+        :metadata,
+        :_private_attribute,
```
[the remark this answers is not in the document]
wasnotrice: The reason I created a new view was so that the "basic" tests could keep using the simpler view, and I wanted to make as small of an impact on existing tests as I could. But I agree, this view duplicates a lot. I'd be happy to merge them if that's preferred

lib/jsonapi/utils/underscore.ex
wasnotrice: Well, I guess not—the module attribute doesn't work for our tests, since we change the config at runtime—I made the config into a function instead

lib/jsonapi/utils/underscore.ex
```diff
@@ -60,10 +64,27 @@ defmodule JSONAPI.Utils.Underscore do
...
-    if is_map(value) do
+    if underscore?(key) and is_map(value) do
       {underscore(key), underscore(value)}
     else
       {underscore(key), value}
     end
   end
+
+  defp underscore?(key) do
+    config_specifies_underscore?(config(), key)
```
snewcomer: could we unwrap this and use `config_specifies_underscore?` directly?

snewcomer:
Lots of really nice work here Eric!

One question that I thought I should at least ask, while hopefully not discounting the great work you did - **Do we want this configuration here?**  The only reason I thought I would ask is the original issue pointed out a missing gap with keys with double underscores and what not.  That was fixed in #139.   Essentially wondering - what the **strong** motivation for this?  More or less simply curious as I don't have a need but I'm sure others here do. 👍

wasnotrice:
@snewcomer there is more to the original issue.

The original issue was that my API allows a user to store arbitrary JSON. But when they retrieve the JSON they stored, it get processed by `jsonapi`'s `underscore_to_dash` setting. In this particular case, fixing the regex to ignore leading underscores does work around the issue. However, that's only a coincidence, since incoming keys tend to be camelcased, not underscored.

So, the **strong** motivation for this is to allow people to control the format of their jsonapi keys (underscore/dash) while also allowing them to designate content whose format **should not be changed, regardless of the format of the response keys.** To that end, I think the most effective place for this is not actually in the application env, but instead per-view. I think that's totally doable, and if folks agree, I could build that out either here, or in a followup PR.

I opened this PR in order to provide a more concrete way to discuss the feature, so discussing it here is totally appropriate! Although I guess it would make more sense to focus first on this sort of high-level question of appropriateness/desirability before we get deep into implementation details :)

lib/jsonapi/utils/underscore.ex
```diff
@@ -42,9 +42,13 @@ defmodule JSONAPI.Utils.Underscore do
       "_top__posts_"
   """
   def underscore(value) when is_atom(value) do
```
snewcomer: These method names seem confusing to me :). Really it is `dasherize`

snewcomer:
I think generally I am hoping to get some more broad thoughts on how `jsonapi` should handle the spec.

From my understanding, JSONAPI defines strong conventions for how the `document` should look (keywords like **MUST, MAY, MUST NOT**).  For example, here is the specification regarding member names and what it can and can't include.

> Additionally, the following characters are allowed in member names, except as the first or last character:
> U+002D HYPHEN-MINUS, “-“
> U+005F LOW LINE, “_”
> U+0020 SPACE, “ “ (not recommended, not URL safe)

https://jsonapi.org/format/#document-member-names

So my question is, what level of "strictness" should we apply?  If we followed the spec, then `_private_attr` would be tossed out.  But that would be too limiting in my opinion.  On the other hand, allowing configuration of `only` might be too loose. i.e. you are implementing a JSONAPI compliant backend but many of your keys are not compliant?  `except` seems to me to provide some leniency, especially when integrating multiple data sources.  On the other hand, you then can put non URL safe characters in there (e.g. `%`).  

There are escape hatches such as `meta` where you can put non standard information in, but that would require some finagling when an easier option would be `except` - https://jsonapi.org/format/#document-meta

All of this is simply discussion points and not meant to discredit any ideas here.  Would love to hear some more thoughts from more ppl!

lib/jsonapi/utils/underscore.ex
wasnotrice: :) yes I agree, they would make more sense to me if switched

doomspork:
Sorry @snewcomer I hadn't seen your comments here.  This is a very valid discussion and I want to mull this over a bit before making a decision.  I'd love to get @jeregrine's thoughts as well 😁

jherdman:
@wasnotrice I was wondering if you're still interested in submitting this PR. We'd love to include it in a future release.

wasnotrice:
@jherdman absolutely—I'd love to get it in when we decide how it fits into the project!

PR #139, opened 2018-11-01
Update regexes for underscore and dash

wasnotrice:
Only match a single underscore/dash in the middle of word characters.
Makes the replacement less zealous.

Extracted from #137

### DIFF
--- a/lib/jsonapi/utils/underscore.ex
+++ b/lib/jsonapi/utils/underscore.ex
@@ -5,10 +5,42 @@ defmodule JSONAPI.Utils.Underscore do
 
   def underscore?, do: Application.get_env(:jsonapi, :underscore_to_dash, false)
 
+  @doc """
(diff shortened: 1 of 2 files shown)


snewcomer:
@wasnotrice Nice! What do you think about moving this test to over to here? 

https://github.com/jeregrine/jsonapi/pull/137/files#diff-b30adaf0c44cc583c847765481e1acf8R94

wasnotrice:
@snewcomer I agree with you about adding a test. Of course there isn't actually a test in the suite for the regexes—they only get tested indirectly through query parser and serializer tests.

I added some doctests for the `underscore/1` and `dash/1` functions. I think this is a nice addition to the overall test suite and documentation. What do you think?

doomspork:
Thank you @wasnotrice!  😁

PR #140, opened 2018-11-02
Improve Application.env handling in tests

wasnotrice:
In tests that require making changes to the application env, the changes do not currently get cleaned up in the case of a test failure, which causes subsequent tests to run with the modified application env, and usually causes surprising failures.

This PR wraps the env-changing tests in a describe block, which has an `on_exit` block that ensures the application env gets reset even when tests fail.


wasnotrice:
This is the first commit in #137. Merging this first will simplify review for #137

PR #146, opened 2018-12-19
Consolidate Plug Locations

jherdman:
Move these two plugs with their other friends.


doomspork:
I was mostly referring to `JSONAPI.PlugResponseContentType`, the `Plug` prefix is unnecessary. If we're going to move the files now would be the time to fix module names otherwise there's not much reason to move things around at this time, it's just churn.


> I'm assuming you want something like JSONAPI.Plugs.MyPlug.

I personally would for clarity, but I'd like to hear what @jeregrine and @snewcomer think

> If this is the case, is it possible to alias module names so as not to break existing code?

I don't think we want to start creating alias for backwards compatibility, maintaining that long term would be a nightmare. It's possible to cut a new release with a new version to signify breaking changes.

jherdman:
> I don't think we want to start creating alias for backwards compatibility, maintaining that long term would be a nightmare. It's possible to cut a new release with a new version to signify breaking changes.

Agreed. It'd be nice to do this if even just to give people a fair heads up.

> I was mostly referring to JSONAPI.PlugResponseContentType, the Plug prefix is unnecessary...

Fixing!

jherdman:
@doomspork after some silence on the matter I decided to make the changes you requested.

test/jsonapi/plugs/response_content_type_test.exs
```diff
@@ -0,0 +1,27 @@
+defmodule JSONAPI.ResponseContentTypeTest do
```
jherdman: @doomspork FYI added a test for this plug too, it was missing one.

doomspork:
@jherdman sorry! The run up the holidays is hectic around here.

PR #149, opened 2018-12-24
Add Plug to Transform Parameters

### LINKED ISSUE #145: Should Filter Params Be Transformed?
Maybe I'm writing by tests wrong, but it seems like filter params aren't transformed. E.g. given...

```elixir
  plug JSONAPI.QueryParser,
    filter: ~w[dog-breed],
    view: MyApp.DogView
```

```
GET /api/foo?filter[dog-breed]="corgi"
```

Results in `%{"filter" => %{"dog-breed" => "corgi"}}`.

It seems to me that we should apply the underscore transformation a la attribute params. This would mean that we neither need to apply the transformation manually, nor does the client need KNOW how to write the params (e.g. if we're dasherized, we're always dasherized).

I know the filter portion of the spec is skimpy, but I figured I'd raise this. Thoughts?

jherdman:
This, optional, Plug can be used to transform incoming dasherized
parameters to underscored ones that are more amenable to usage in your
application.

Resolves #145

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,9 @@ A project that will render your data models into [JSONAPI Documents](http://json
 
 ## JSONAPI Support
 
+This library implements [version 1.0](https://jsonapi.org/format/1.0/)
+of the JSON:API spec.
+
 - [x] Basic [JSONAPI Document](http://jsonapi.org/format/#document-top-level) encoding
 - [x] Basic support for [compound documents](http://jsonapi.org/format/#document-compound-documents)
 - [x] [Links](http://jsonapi.org/format/#document-links)
@@ -19,7 +22,8 @@ A project that will render your data models into [JSONAPI Documents](http://json
 
 ## Documentation
 
-[Full docs here](https://hexdocs.pm/jsonapi/api-reference.html)
+* [Full docs here](https://hexdocs.pm/jsonapi)
+* [JSON API Spec (v1.0)](https://jsonapi.org/format/1.0/)
 
 ## How to use with Phoenix
 
@@ -80,6 +84,30 @@ when Ecto gets more complex field selection support we will go further to only q
 
 You will need to handle filtering yourself, the filter is just a map with key=value.
 
+## Dasherized Fields
+
+JSONAPI now recommends the use of dashes (`-`) in place of underscore (`_`) as a
+word separator. Handling these fields requires two steps:
+
+1. Dasherizing *outgoing* fields requires you to set the `underscore_to_dash`
+   configuration option. This is not enabled by default. Example:
+
+    ```elixir
+    config :jsonapi,
+      underscore_to_dash: true,
+    ```
+
+2. Underscoring *incoming* params (both query and body) requires you add the
+   `JSONAPI.UnderscoreParameters` Plug to your API's pipeline. Your pipeline in a
+   Phoenix app might look something like this:
+
+   ```elixir
+   pipeline :api do
+     plug(JSONAPI.EnsureSpec)
+     plug(JSONAPI.UnderscoreParameters)
+   end
+   ```
+
 ## Spec Enforcement
 
 We include a set of Plugs to make enforcing the JSONAPI spec for requests easy. To add spec enforcement to your application, add `JSONAPI.EnsureSpec` to your pipeline:
@@ -98,8 +126,6 @@ Under-the-hood `JSONAPI.EnsureSpec` relies on three individual plugs:
 
 ## Configuration
 
-By default host and scheme are pulled from the provided conn but can be overridden via configuration like so:
-
 ```elixir
 config :jsonapi,
   host: "www.someotherhost.com",
@@ -109,13 +135,12 @@ config :jsonapi,
   json_library: Jason
 ```
 
-- **underscore_to_dash**
-
-Additionally, JSONAPI now recommends the use of dashes (`-`) in place of underscore (`_`) as a word separator. Enabling this change is easy with the `underscore_to_dash` option, which handles the conversion for you. Defaults to `false`.
-
-- **remove_links**
-
-`links` data can optionally be removed from the payload via setting the configuration above to `true`. Defaults to `false`.
+- **host**, **scheme**. By default these are pulled from the `conn`, but may be
+  overridden.
+- **remove_links**. `links` data can optionally be removed from the payload via
+  setting the configuration above to `true`. Defaults to `false`.
+- **json_library**. Defaults to [Jason](https://hex.pm/packages/jason).
+- **underscore_to_dash**. See "Dasherizing Fields".
 
 ## Other
 
--- a/lib/jsonapi/deprecation.ex
+++ b/lib/jsonapi/deprecation.ex
@@ -20,4 +20,11 @@ defmodule JSONAPI.Deprecation do
       Macro.Env.stacktrace(__ENV__)
     )
   end
+
+  def warn(:query_parser_dash) do
+    IO.warn(
+      "`JSONAPI.QueryParser` will no longer automatically dasherize incoming parameters. Please include `JSONAPI.UnderscoreParameters` in your pipeline. See https://github.com/jeregrine/jsonapi/pull/149",
+      Macro.Env.stacktrace(__ENV__)
+    )
+  end
 end
--- a/lib/jsonapi/plugs/query_parser.ex
+++ b/lib/jsonapi/plugs/query_parser.ex
@@ -58,6 +58,14 @@ defmodule JSONAPI.QueryParser do
     * `:view` - The JSONAPI View which is the basis for this plug.
     * `:sort` - List of atoms which define which fields can be sorted on.
     * `:filter` - List of atoms which define which fields can be filtered on.
+
+  ## Dasherized Fields
+
+  Note that if your API is returning dasherized params (e.g. `"dog-breed": "Corgi"`)
+  we recommend that you include the `JSONAPI.UnderscoreParameters` Plug in your
+  API's pipeline. This will underscore fields for easier operations in your code.
+
+  For more details please see `JSONAPI.UnderscoreParameters`.
   """
 
   def init(opts) do
--- a/lib/jsonapi/plugs/query_parser.ex
+++ b/lib/jsonapi/plugs/query_parser.ex
@@ -217,6 +217,7 @@ defmodule JSONAPI.QueryParser do
 
   def dash(data) do
     if Underscore.underscore?() do
+      Deprecation.warn(:query_parser_dash)
       Underscore.dash(data)
     else
       data
--- a/lib/jsonapi/plugs/underscore_parameters.ex
+++ b/lib/jsonapi/plugs/underscore_parameters.ex
@@ -30,7 +30,7 @@ defmodule JSONAPI.UnderscoreParameters do
   def call(%Plug.Conn{params: params} = conn, _opts) do
     content_type = get_req_header(conn, "content-type")
 
-    if @json_api_content_type in content_type do
+    if JSONAPI.mime_type() in content_type do
       new_params = dash(params)
 
       Map.put(conn, :params, new_params)
--- a/lib/jsonapi/plugs/underscore_parameters.ex
+++ b/lib/jsonapi/plugs/underscore_parameters.ex
@@ -1,0 +1,41 @@
+defmodule JSONAPI.UnderscoreParameters do
+  @moduledoc """
+  Takes dasherized JSON:API params and turns them into underscored params. Add
+  this to your API's pipeline to aid in dealing with incoming parameters.
+
+  ## Example
+
+  Given a request like:
+
+      GET /example?filters[dog-breed]=Corgi
+
+  You could implement your corresponding `index` method like:
+
+      def index(conn, %{"filters" => %{"dog_breed" => "Corgi"}})
+
+  Without this Plug your index action would look like:
+
+      def index(conn, %{"filters" => %{"dog-breed" => "Corgi"}})
+  """
+
+  import Plug.Conn
+
+  import JSONAPI.Utils.Underscore, only: [dash: 1]
+
+  @doc false
+  def init(_opts) do
+  end
+
+  @doc false
+  def call(%Plug.Conn{params: params} = conn, _opts) do
+    content_type = get_req_header(conn, "content-type")
+
+    if @json_api_content_type in content_type do
+      new_params = dash(params)
+
+      Map.put(conn, :params, new_params)
+    else
+      conn
+    end
+  end
+end
--- a/lib/jsonapi/plugs/underscore_parameters.ex
+++ b/lib/jsonapi/plugs/underscore_parameters.ex
@@ -3,19 +3,32 @@ defmodule JSONAPI.UnderscoreParameters do
   Takes dasherized JSON:API params and turns them into underscored params. Add
   this to your API's pipeline to aid in dealing with incoming parameters.
 
+  Note that this Plug will only underscore parameters when the request's content
+  type is for a JSON:API request (i.e. "application/vnd.api+json"). All other
+  content types will be ignored.
+
   ## Example
 
   Given a request like:
 
       GET /example?filters[dog-breed]=Corgi
 
-  You could implement your corresponding `index` method like:
+  **Without** this Plug your index action would look like:
+
+      def index(conn, %{"filters" => %{"dog-breed" => "Corgi"}})
+
+  And **with** this Plug:
 
       def index(conn, %{"filters" => %{"dog_breed" => "Corgi"}})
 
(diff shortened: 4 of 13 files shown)


test/jsonapi/plugs/transform_parameters_test.exs
```diff
@@ -0,0 +1,68 @@
...
+          }
+        }
+      },
+      "filter" => %{
+        "dog-breed" => "Corgi"
+      }
+    }
+
+    conn =
+      conn(:get, "/hello", params)
```
doomspork: Please start pipes with a bare value:

```elixir
:get
|> conn("/hello", params)
|> put_req_header("content-type", "application/vnd.api+json")
```

lib/jsonapi/plugs/transform_parameters.ex
```diff
@@ -0,0 +1,41 @@
...
+  def init(opts), do: opts
+
+  def call(%Plug.Conn{params: params} = conn, _opts) do
+    content_type = get_req_header(conn, "content-type")
+
+    if @json_api_content_type in content_type do
+      new_params = dasherize_keys(params)
+
+      conn
+      |> Map.put(:params, new_params)
```
doomspork: We don't need to use a pipe here.

lib/jsonapi/utils/underscore.ex
```diff
@@ -25,6 +25,31 @@ defmodule JSONAPI.Utils.Underscore do
...
+
+  ## Examples
+
+      iex> dasherize_keys(%{"foo-bar" => "baz"})
+      %{"foo_bar" => "baz"}
+
+      iex> dasherize_keys(%{"f-b" => %{"a-d" => "z"}, "c-d" => "e"})
+      %{"f_b" => %{"a_d" => "z"}, "c_d" => "e"}
+  """
+  def dasherize_keys(map) when is_map(map) do
```
doomspork: We don't need the guards on `dasherize_keys/1` and `dash/1` if we only use them for specific values. What do you think about using `dash/1` and pattern matching to do our work?

```elixir  
@doc """
  Replace dashes between words in `value` with underscores

  Ignores dashes that are not between letters/numbers

  ## Examples

      iex> dash("top-posts")
      "top_posts"

      iex> dash("-top-posts")
      "-top_posts"

      iex> dash("-top--posts-")
      "-top--posts-"

      iex> dash(%{"foo-bar" => "baz"})
      %{"foo_bar" => "baz"}

      iex> dash({"foo-bar", "baz"})
      {"foo_bar", "baz"}
  """
  def dash(value) when is_binary(value) do
    String.replace(value, ~r/([a-zA-Z0-9])-([a-zA-Z0-9])/, "\\1_\\2")
  end

  def dash(map) when is_map(map) do
    Enum.reduce(map, %{}, &dash/1)
  end

  def dash({key, value}) do
    {dash(key), dash(value)}
  end

  def dash(value) do
    value
  end
```

doomspork: We could also handle atoms though this is really not something we should encourage:

```elixir
def dash(value) when is_atom(value) do
  value
  |> to_string()
  |> dash()
  |> String.to_atom()
end
```

jherdman: Yeah, I'm into that! I'll make that change.

jherdman:
TODO
------

- [ ] Add documentation to README
- [ ] Resolve all feedback

lib/jsonapi/plugs/transform_parameters.ex
```diff
@@ -0,0 +1,41 @@
+defmodule JSONAPI.TransformParameters do
```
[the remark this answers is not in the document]
doomspork: I think a better name is a good idea, `Transform` is rather generic

lib/jsonapi/utils/underscore.ex
jherdman: I'm 100% into renaming these functions. This was a source of confusion for me for sure!

lib/jsonapi/plugs/transform_parameters.ex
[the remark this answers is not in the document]
jherdman: One reason I kept the generic name is so that if we eventually want to add functions to handle camelCase we're in the clear. Come to think of it, though, we may want a distinct plug to deal with such a scenario.

lib/jsonapi/plugs/transform_parameters.ex
```diff
@@ -0,0 +1,41 @@
...
+
+  @json_api_content_type "application/vnd.api+json"
+
+  def init(opts), do: opts
+
+  def call(%Plug.Conn{params: params} = conn, _opts) do
+    content_type = get_req_header(conn, "content-type")
+
+    if @json_api_content_type in content_type do
+      new_params = dasherize_keys(params)
```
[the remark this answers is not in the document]
jherdman: My original intention was to just handle filter params, but I think we can consolidate down into a single Plug. I still haven't removed the duplication present in the query parser plug, and I _think_ we ought to (though this will be a breaking change if the plug introduced in this PR remains optional unless we can bundle them???).

lib/jsonapi/utils/underscore.ex
jherdman: @snewcomer is part of the intention of #137 to rename these methods? If not I'll open a PR to do so.

lib/jsonapi/utils/underscore.ex
```diff
@@ -20,10 +20,39 @@ defmodule JSONAPI.Utils.Underscore do
...
+      %{"f_b" => %{"a_d" => "z"}, "c_d" => "e"}
+
+      iex> dash(:"foo-bar")
+      :foo_bar
   """
   def dash(value) when is_binary(value) do
     String.replace(value, ~r/([a-zA-Z0-9])-([a-zA-Z0-9])/, "\\1_\\2")
   end
+  def dash(map) when is_map(map) do
+    Enum.reduce(map, %{}, fn ({k, v}, acc) ->
```
jherdman: Is there a more elegant way to write this? I tried `Enum.reduce(map, %{}, &dash/1)`, but since the dashing function doesn't accept an accumulator things would go sideways. It seemed silly to add a `dash/2` just to support this as well.

jherdman: Done! That's pretty darn baller.

test/jsonapi/plugs/transform_parameters_test.exs
```diff
@@ -0,0 +1,69 @@
+defmodule DailyBugleWeb.TransformParametersTest do
```
doomspork: I think this namespace is a mistake 😁 

Shall we add doctest to this as well since we have it setup in the module?

lib/jsonapi/plugs/transform_parameters.ex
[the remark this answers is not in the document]
doomspork: Let's change the name to be more specific for now. If we decide in the future to support additional functionality, we can cross that bridge then.

test/jsonapi/plugs/transform_parameters_test.exs
jherdman: > I think this namespace is a mistake

D'oh! Wanna guess what the code name of my side project is? ;)

> Shall we add doctest to this as well since we have it setup in the module?

Yeah, may as well!

jherdman:
> Overall a user has to understand, to underscore params, add a plug. To underscore a response object, add a config option...

This asymmetry kind of stinks. Maybe we could optionally include the plug with the config option set?

lib/jsonapi/plugs/underscore_parameters.ex
```diff
@@ -0,0 +1,40 @@
...
+  Without this Plug your index action would look like:
+
+      def index(conn, %{"filters" => %{"dog-breed" => "Corgi"}})
+  """
+
+  import Plug.Conn
+
+  import JSONAPI.Utils.Underscore, only: [dash: 1]
+
+  @json_api_content_type "application/vnd.api+json"
```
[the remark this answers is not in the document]
jherdman: OK, I did this in a new commit. Seemed too obvious a refactor.

test/jsonapi/plugs/transform_parameters_test.exs
jherdman: I tried to set up some doc tests, but it was quite awkward.

test/jsonapi/plugs/content_type_negotiation_test.exs
```diff
@@ -178,14 +180,14 @@ defmodule JSONAPI.ContentTypeNegotiationTest do
...
-        "application/vnd.api+json; version=1.0, application/vnd.api+json; version=1.0"
+        "#{mime_type()}; version=1.0, #{mime_type()}; version=1.0"
       )
       |> ContentTypeNegotiation.call([])
 
     assert conn.halted
 
     assert Plug.Conn.get_resp_header(conn, "content-type") == [
-             "application/vnd.api+json; charset=utf-8"
+             "#{mime_type()}; charset=utf-8"
```
snewcomer: nice cleanup 👍

README.md
```diff
@@ -109,13 +107,36 @@ config :jsonapi,
...
+- **host**, **scheme**. By default these are pulled from the `conn`, but may be
+  overridden.
+- **remove_links**. `links` data can optionally be removed from the payload via
+  setting the configuration above to `true`. Defaults to `false`.
+- **json_library**. Defaults to [Jason](https://hex.pm/packages/jason).
+- **underscore_to_dash**. See "Dasherizing Links".
+
+## Dasherizing Fields
+
+JSONAPI now recommends the use of dashes (`-`) in place of underscore (`_`) as a
```
snewcomer: note this is now camelCase :trollface: But that will happen in a follow up PR.

lib/jsonapi/utils/underscore.ex
```diff
@@ -20,10 +20,37 @@ defmodule JSONAPI.Utils.Underscore do
...
+  end
+  def dash(value) when is_atom(value) do
+    value
+    |> to_string()
+    |> dash()
+    |> String.to_atom()
+  end
+  def dash(value) do
+    value
+  end
```
snewcomer: following up on the previous conversation, I think the swap of method names could come in another PR or in #137.  This looks great as is!

README.md
```diff
@@ -109,13 +107,36 @@ config :jsonapi,
...
+word separator. Enabling this requires two quick steps:
+
+1. Dasherizing *outgoing* fields requires you to set the `underscore_to_dash`
```
snewcomer: nice!

README.md
```diff
@@ -109,13 +107,36 @@ config :jsonapi,
...
+   configuration option. This is not enabled by default. Example:
 
-Additionally, JSONAPI now recommends the use of dashes (`-`) in place of underscore (`_`) as a word separator. Enabling this change is easy with the `underscore_to_dash` option, which handles the conversion for you. Defaults to `false`.
+    ```elixir
+    config :jsonapi,
+      underscore_to_dash: true,
+    ```
 
-- **remove_links**
+2. Dasherizing *incoming* fields requires you add the `JSONAPI.UnderscoreParamters`
```
snewcomer: perhaps clarify this underscores filter params **and** body params.

jherdman:
I'm pretty sure this is in a "done" state now.

I'd like to recommend that a breaking 1.0 release project be opened as a result of this PR. Maybe we should sneak in a few other breaking changes whilst we're at it?

snewcomer:
@jherdman Nice work!  Sry for asking, but what is the breaking change here?  This just adds functionality, right?  I most likely missed something.

Also, I think there are a few items we **could** think about for 1.0.

1. `profile` links https://jsonapi.org/format/1.1/.  (~January 2019) albeit non breaking
2. camelCase member names (if we decide to switch over to).  This is totally up in the air and we could stick with conventions in the Elixir community

README.md
```diff
@@ -19,7 +19,7 @@ A project that will render your data models into [JSONAPI Documents](http://json
 
 ## Documentation
 
-[Full docs here](https://hexdocs.pm/jsonapi/api-reference.html)
+[Full docs here](https://hexdocs.pm/jsonapi)
```
doomspork: Should we include a link to the actual JSONAPI spec here?

lib/jsonapi/plugs/transform_parameters.ex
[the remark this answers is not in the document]
snewcomer: @jherdman I think opt in behavior would be a minor bump.  Is this an incompatible API change for all users?  I think this is technically backwards compat with added functionality.  Feel free to correct me!

README.md
jherdman: Done!

lib/jsonapi/plugs/query_parser.ex
```diff
@@ -179,7 +186,7 @@ defmodule JSONAPI.QueryParser do
       if inc =~ ~r/\w+\.\w+/ do
         acc ++ handle_nested_include(inc, valid_includes, config)
       else
-        inc = inc |> dash() |> String.to_existing_atom()
+        inc = String.to_existing_atom(inc)
```
jherdman: @snewcomer here's an example of a backwards incompat change. I could add it back in with a deprecation warning.

README.md
```diff
@@ -80,6 +84,30 @@ when Ecto gets more complex field selection support we will go further to only q
...
+1. Dasherizing *outgoing* fields requires you to set the `underscore_to_dash`
+   configuration option. This is not enabled by default. Example:
+
+    ```elixir
+    config :jsonapi,
+      underscore_to_dash: true,
+    ```
+
+2. Underscoring *incoming* params (both query and body) requires you add the
+   `JSONAPI.UnderscoreParamters` Plug to your API's pipeline. Your pipeline in a
```
snewcomer: r/Paramters/Parameters

lib/jsonapi/plugs/query_parser.ex
snewcomer: Since we are doing the whole shebang, probably ok.

README.md
```diff
@@ -80,6 +84,30 @@ when Ecto gets more complex field selection support we will go further to only q
 
 You will need to handle filtering yourself, the filter is just a map with key=value.
 
+## Dasherized Fields
+
+JSONAPI now recommends the use of dashes (`-`) in place of underscore (`_`) as a
```
snewcomer: I'd be curious as to @doomspork or @jeregrine opinion what the library `case` default should be and if we should take into consideration camelCase recommendation now.  Or should we stick with general Elixir conventions for underscore and leave as it, providing configuration on top to convert to `dash` or `camelCase` based on what the consumer wants?

jherdman:
Reverted the hard deprecation on `QueryParser` dasherizing. This wasn't a lot of fun when I tried this branch out in my project.

README.md
jherdman: As someone using this lib I'd prefer to opt into whatever I want, even on a request level. Imagine an API like Stripe's where you tell the API what version of it you're expecting. In such a case I'd want to either dasherize for old clients, or camelize for modern ones. FWIW as a default I'd love to see at least dashes given that this is the v1.0 recommendation.

jherdman:
Initial testing has show that this also underscore values, e.g. if I `POST /sources` with the payload like this:

```
{
  "data": {
    "attributes": {
      "math-problem": "1-1=2"
```

I'll get:

```
{
  "data": {
    "attributes": {
      "math-problem": "1_1=2"
```

jherdman:
I've beefed up the tests to account for the above problems, and am successfully using this branch in a project. I'm reasonably sure this is good to go unless there's any more feedback.

doomspork:
@jherdman can you please undo the change `LICENSE → LICENSE.md`, we do not need to add a `.md` extension. There's really no reason to make these changes as part of this PR. Please _please_ keep PRs focused on the work you're doing and avoid unnecessary changes 👍

jherdman:
The reason for this change is to include it in the docs. ExDoc gets cranky
otherwise.

Apologies about this PR ballooning.

jherdman:
@doomspork reverted!

doomspork:
@snewcomer any final thoughts before I merge this? I believe we will need @jeregrine to cut a release once we have everything merged and ready.

lib/jsonapi/deprecation.ex
```diff
@@ -20,4 +20,11 @@ defmodule JSONAPI.Deprecation do
       Macro.Env.stacktrace(__ENV__)
     )
   end
+
+  def warn(:query_parser_dash) do
+    IO.warn(
+      "`JSONAPI.QueryParser` will no longer automatically dasherize incoming parameters. Please include `JSONAPI.UnderscoreParameters` in your pipeline. See https://github.com/jeregrine/jsonapi/pull/149",
```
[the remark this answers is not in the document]
doomspork: Let's do that in a separate PR so @jherdman can get this off his plate.

PR #150, opened 2018-12-28
Expand Build Matrix

jherdman:
Let's also trust Travis to do the right thing with respect to selecting
the OTP version to use.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,10 @@
 language: elixir
 
 elixir:
+  - 1.7
   - 1.6
   - 1.5
   - 1.4
-
-otp_release:
-  - 20.1
-  - 19.3
-  - 18.3
-
-matrix:
-  exclude:
-    - elixir: 1.6
-      otp_release: 18.3
 
 env:
   global:
@@ -27,4 +18,4 @@ sudo: false
 script:
   - mix test
   - mix credo
-  - if [[ `elixir -v` = *"1.6"* ]]; then mix format --check-formatted; fi
+  - if [[ `elixir -v` = *"1.7"* ]]; then mix format --check-formatted; fi


.travis.yml
```diff
@@ -27,4 +18,4 @@ sudo: false
 script:
   - mix test
   - mix credo
-  - if [[ `elixir -v` = *"1.6"* ]]; then mix format --check-formatted; fi
+  - if [[ `elixir -v` = *"1.7"* ]]; then mix format --check-formatted; fi
```
doomspork: Why did you change this to 1.7?

.travis.yml
```diff
@@ -1,20 +1,11 @@
 language: elixir
 
 elixir:
+  - 1.7
   - 1.6
   - 1.5
   - 1.4
 
-otp_release:
```
doomspork: Can you link us to the documentation that explains why we don't need this? If you look around pretty much every Elixir project has these OTP releases defined, including Elixir itself.

jherdman: https://docs.travis-ci.com/user/languages/elixir#specifying-otp-release-version I think I'm reading this correctly...

.travis.yml
jherdman: 1.7, being the latest public release (outside of 1.8-rc1), felt like the logical choice to run this particular build step on — especially since I'm adding it to the build.

.travis.yml
doomspork: I believe so. As I understand it, if we leave this off, it will pick _one_ OTP version. We are intentionally testing against multiples because it is common in real-world situations to have Elixir and OTP on different versions.

.travis.yml
doomspork: This is meant to ensure the command isn't run without formatter (versions prior to 1.5).

doomspork:
@jherdman I'm going to close this PR for now. None of these changes will work as expected:

1) We intentionally want to test against multiple OTP versions, not let Travis pick just one.
2) The 1.6 guard is to ensure the formatter is only run for versions in which it exists (so not 1.3, 1.4, and 1.5). It should not be updated with each new release.


Next time feel free to open an issue first to discuss changes not currently documented 👍

PR #151, opened 2018-12-29
Expand build matrix

doomspork:
Going the opposite way of #150 by actually expanding the matrix to include the latest Elixir and OTP releases.  As mentioned on #150, we intentionally want multiple OTP releases so we can ensure the project compiles on different combinations of Elixir and OTP.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,13 @@
 language: elixir
 
 elixir:
+  - 1.7
   - 1.6
   - 1.5
-  - 1.4
 
 otp_release:
-  - 20.1
+  - 21.1
+  - 20.3
   - 19.3
   - 18.3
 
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,19 +1,22 @@
...
+  - 20.3
   - 19.3
   - 18.3
 
 matrix:
   exclude:
     - elixir: 1.6
       otp_release: 18.3
+    - elixir: 1.5
+      otp_release: 21.1
```
jherdman: I think you need to add an exclusion for 1.7 and OTP < 19 here too. See https://hexdocs.pm/elixir/compatibility-and-deprecations.html#compatibility-between-elixir-and-erlang-otp

doomspork: Are you sure? All the tests currently pass @jherdman

doomspork: image

.travis.yml
```diff
@@ -1,19 +1,22 @@
...
 
 env:
   global:
```
snewcomer: I think `sudo: false` can be removed now.  https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration

PR #152, opened 2018-12-30
Move `QueryParser` Test

jherdman:
In #146 the `QueryParser` Plug was moved alongside its friends.
Unfortunately its test wasn't moved too. Let's fix that.


jherdman:
@doomspork apologies for not opening an issue first, I felt like this one was a no brainer given #146, and this really should have been done there too.

doomspork:
Thanks @jherdman, this is good 👍

PR #155, opened 2019-01-09
Refactor String Manipulation Utility Module

### LINKED ISSUE #153: [RFC] Tweak Utils.Underscore Module
Proposal
---------

* `Utils.Underscore` module should be renamed `Utils.String` — it exclusively operates on Strings and this new name makes way for eventually offering camelization and decamelization without a confusing name.
* `Utils.Underscore.dash/1` and `Utils.Underscore.underscore/1` should change names. The current names are confusing: `dash/1` replaces dashes with underscores, and `underscore/1` replaces dashes with underscores. Other libraries (e.g. Rails, Ember) also use `dasherize`. Aligning with these libraries eases the barrier to entry for new users.

Concerns
-----------

* This is a breaking change to a public API
* Merge conflicts for WIP PRs

Possible Mitigation Pathways
-----------------------------

Introduce a module using the proposed named, `Utils.String`, and use the new method names. The current module, `Utils.Underscore`, will point to these methods and include a deprecation warning. This would allow us to safely do a minor release.

Alternatives
------------

### Replaces `Utils.Underscore` with Inflex

[Inflex](https://github.com/nurugger07/inflex) is a library that offers String manipulation functions that could replace _part of_:

* `Utils.Underscore.underscore` with `Inflex.dash`
* `Utils.Underscore.dash` with `Inflex.parameterize(term, "-")`
* And even supplies `Inflex.camelize`

We could replace the String manipulation functions in `Utils.Underscore` with said methods, leaving in place utilities for traversing objects (e.g. changes added in #149 to traverse params and underscore them). Said utilities could be moved to a `Utils.Parameters` module.

#### Advantages

* Inflex is fleshed out and tested. We wouldn't have to maintain such functions going forward.
* Above mitigation pathway still applies

#### Concerns

* We lose control over these methods. Future changes with Inflex may not map to our needs, or even what we need exactly.
* Above concerns still apply

#### Open Questions

I'll try to get answers for these ASAP:

- [x] Are Inflex's transformations 100% compatible with our own? E.g. does `underscore/1` work _exactly_ like `parameterize(term, "-")`?
- [x] License concerns?
- [x] Similar libraries that we should investigate?

jherdman:
* Introduces `Utils.StringManipulation` for holding any String
  manipulation functions
* Renames functions to have a more intuitive name
* Deprecates old module and corresponding functions
* Rolls in "safe" String manipulation methods into the new module

Deviations from RFC
-------------------

I ended up using the module name `StringManipulation` instead of `String`. I quickly ran into issues when trying to alias using the proposed name.

Resolves #153

### DIFF
--- a/lib/jsonapi/plugs/query_parser.ex
+++ b/lib/jsonapi/plugs/query_parser.ex
@@ -2,8 +2,8 @@ defmodule JSONAPI.QueryParser do
   @behaviour Plug
   alias JSONAPI.{Config, Deprecation, Page}
   alias JSONAPI.Exceptions.InvalidQuery
+  alias JSONAPI.Utils.String, as: JString
   alias Plug.Conn
-  alias JSONAPI.Utils.Underscore
   import JSONAPI.Utils.IncludeTree
 
   @moduledoc """
@@ -61,7 +61,7 @@ defmodule JSONAPI.QueryParser do
 
   ## Dasherized Fields
 
-  Note that if your API is returning dasherized params (e.g. `"dog-breed": "Corgi"`)
+  Note that if your API is returning dasherized fields (e.g. `"dog-breed": "Corgi"`)
   we recommend that you include the `JSONAPI.UnderscoreParameters` Plug in your
   API's pipeline. This will underscore fields for easier operations in your code.
 
@@ -187,7 +187,7 @@ defmodule JSONAPI.QueryParser do
       if inc =~ ~r/\w+\.\w+/ do
         acc ++ handle_nested_include(inc, valid_includes, config)
       else
-        inc = inc |> dash() |> String.to_existing_atom()
+        inc = inc |> normalize_fields() |> String.to_existing_atom()
 
         if Enum.any?(valid_includes, fn {key, _val} -> key == inc end) do
           acc ++ [inc]
@@ -202,7 +202,7 @@ defmodule JSONAPI.QueryParser do
     keys =
       key
       |> String.split(".")
-      |> Enum.map(&dash/1)
+      |> Enum.map(&normalize_fields/1)
       |> Enum.map(&String.to_existing_atom/1)
 
     last = List.last(keys)
@@ -212,15 +212,6 @@ defmodule JSONAPI.QueryParser do
       put_as_tree([], path, last)
     else
       raise InvalidQuery, resource: config.view.type(), param: key, param_type: :include
-    end
-  end
-
-  def dash(data) do
-    if Underscore.underscore?() do
-      Deprecation.warn(:query_parser_dash)
-      Underscore.dash(data)
-    else
-      data
     end
   end
 
@@ -254,5 +245,15 @@ defmodule JSONAPI.QueryParser do
       end
 
     struct(struct, processed_map)
+  end
+
+  defp normalize_fields(fields) do
+    Deprecation.warn(:query_parser_dash)
+
+    if JString.field_transformation() == :underscore do
+      fields
+    else
+      JString.underscore(fields)
+    end
   end
 end
--- a/lib/jsonapi/utils/underscore.ex
+++ b/lib/jsonapi/utils/underscore.ex
@@ -1,110 +1,25 @@
 defmodule JSONAPI.Utils.Underscore do
   @moduledoc """
-  Helpers for replacing underscores with dashes.
+  DEPRECATED. Please Use `JSONAPI.Utils.String` instead.
   """
-
-  def underscore?, do: Application.get_env(:jsonapi, :underscore_to_dash, false)
 
   @doc """
   Replace dashes between words in `value` with underscores
 
   Ignores dashes that are not between letters/numbers
-
-  ## Examples
-
-      iex> dash("top-posts")
-      "top_posts"
-
-      iex> dash("-top-posts")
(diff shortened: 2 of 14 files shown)


lib/jsonapi/utils/string_manipulation.ex
```diff
@@ -0,0 +1,148 @@
...
+      # Outgoing fields should be underscored, this is the default
+      config :jsonapi,
+        underscore_to_dash: false
+  """
+  def dasherize?, do: Application.get_env(:jsonapi, :underscore_to_dash, false)
+
+  @doc """
+  Dasherizes a String if, and only if, `dasherize?/0` is true
+  """
+  def maybe_dasherize(term) do
```
jherdman: I'm a bit undecided on rolling these methods in. They feel awfully private to me, or like they shouldn't exist at all. I may pull them back into their old modules and mark them as private.

doomspork: I would prefer we drop prefixes like `maybe_` at the very least.

jherdman: Well, one option we have is to keep at least one of these for now. If we introduce a Plug that transforms OUTGOING params we needn't any of this at all. This would also mean we could have outgoing/incoming pairs for things like having a dasherized or camelized API.

test/utils/string_manipulation_test.exs
```diff
@@ -0,0 +1,53 @@
...
+
+      on_exit(fn ->
+        Application.delete_env(:jsonapi, :underscore_to_dash)
+      end)
+
+      {:ok, []}
+    end
+
+    test "underscores a term" do
+      assert maybe_underscore("foo-bar") == "foo_bar"
```
snewcomer: So if config says keys should be dasherized, then underscore?  Same behaviour on master, but confusing 😆

lib/jsonapi/plugs/query_parser.ex
```diff
@@ -215,15 +215,6 @@ defmodule JSONAPI.QueryParser do
     end
   end
 
-  def dash(data) do
-    if Underscore.underscore?() do
-      Deprecation.warn(:query_parser_dash)
```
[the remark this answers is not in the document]
jherdman: Fixed!

test/utils/string_manipulation_test.exs
jherdman: Should be better now. I've left room for the new option to be used for camelizing.

jherdman:
@snewcomer @doomspork FYI I think I've addressed the last round of feedback if anyone's interested.

lib/jsonapi/utils/underscore.ex
```diff
@@ -1,110 +1,30 @@
...
-    value
-    |> to_string()
-    |> dash()
-    |> String.to_atom()
-  end
-
+  @deprecated "Use JSONAPI.Utils.StringManipulation.underscore/1 instead"
   def dash(value) do
-    value
+    JSONAPI.Utils.StringManipulation.underscore(value)
```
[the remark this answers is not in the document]
doomspork: I'm not too concerned using `alias` because they can always use `:as`. 

I think we should go with `Utils.String` or we could with something like `Utils.Params` since this is a set of utilities for working on the param strings; `StringManipulation` is a handful.

lib/jsonapi/utils/string_manipulation.ex
```diff
@@ -0,0 +1,153 @@
...
+    if value do
+      :dasherize
+    else
+      :underscore
+    end
+  end
+
+  defp normalized_underscore_to_dash_config(value) when is_nil(value), do: value
+
+  @allowed_transformations [:dasherize, :underscore]
```
doomspork: Can we move module attributes to the top of the file?  https://github.com/christopheradams/elixir_style_guide#module-attribute-ordering

jherdman: Done!

lib/jsonapi/utils/underscore.ex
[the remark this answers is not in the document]
jherdman: Done!

lib/jsonapi/plugs/query_parser.ex
```diff
@@ -255,4 +246,14 @@ defmodule JSONAPI.QueryParser do
...
   end
+
+  defp normalize_fields(fields) do
+    Deprecation.warn(:query_parser_dash)
+
+    if JString.field_transformation() == :underscore do
+      fields
+    else
+      JString.underscore(fields)
+    end
```
snewcomer: I'm still unsure what this means.  If `field_transformation` is `:underscore`, then return as it.  Otherwise, underscore the keys?

jherdman: Yeah, it's a tad confusing, but the gist is that if your API is using underscored keys there's no need to underscore them. If your API is using dashes or camel keys, underscore they keys so that we can more easily use them in Elixir.

snewcomer: Ah ok.  That makes sense.  My initial thought was, "hey if you want underscore, let me make sure they are underscore".  But since this predates me, seems ok to leave as it was!

PR #160, opened 2019-01-18
Prepare for 0.9.0

jherdman:
I've updated the change log and bumped the version number in mix.exs.
All we need to do now is cut the release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,26 @@
 # Changelog
 
+## 0.9.0 (2019-01-18)
+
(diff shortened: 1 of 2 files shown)


jherdman:
Let me know if I'm missing anything. If not, merge, cut, and we'll do a little dance!

jherdman:
Question: is it safe to use documentation meta data yet? E.g.

```elixir
@moduledoc since: "0.9.0"
```

If so I could go through an annotate things.

doomspork:
@jherdman I will push to hex 👌

PR #162, opened 2019-01-19
Update ex_doc for release

doomspork:
@jherdman the release fails because we're using a very old `ex_doc`. This should fix that. Could you please review and approve?  Additionally, are you able to tag the release on GitHub? If so please do 😁


doomspork:
This isn't really necessary, I can just update `mix.lock` for the release 👍

PR #164, opened 2019-01-25
Support Arbitrary User-Defined Links

### LINKED ISSUE #141: Links Usage Unclear
I've read over the docs and have done my best to grok the code, but I'm a bit confused on how to specify links.

Suppose I want to include a link for a Book's many Pages, I might want something like this:

```json
{
  "data": {
    "id": 123,
    "title": "Curious George"
  },
  "relationships": {
    "pages": {
      "self": "/api/book/123/relationship/pages"
// snip
```

Setting up my `BookView` isn't quite clear to me. If I override `links/2` it doesn't seem to do anything. Overriding `relationships/0` as so doesn't seem to work:

```elixir
def relationships do
  [pages: PageView]
end
```

I think I may be missing something obvious, but I could use a hand.

jherdman:
JSON:API allows for a `links` object in the `data` payload (see
https://jsonapi.org/format/#document-links). Prior to this commit we
only included user-defined links should they be paginating. This commit
ensures these links are always included unless otherwise specified by
the `:remove_links` option.

Resolves #141

### DIFF
--- a/lib/jsonapi/serializer.ex
+++ b/lib/jsonapi/serializer.ex
@@ -145,6 +145,12 @@ defmodule JSONAPI.Serializer do
     merge_related_links(data, info, remove_links?())
   end
 
+  defp merge_base_links(doc, data, view, conn) do
+    links = Map.merge(%{self: view.url_for(data, conn)}, view.links(data, conn))
+
+    Map.merge(doc, %{links: links})
+  end
+
   defp merge_links(doc, data, view, conn, false, true) do
     pagination_links = view.links(data, conn)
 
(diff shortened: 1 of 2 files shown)


lib/jsonapi/serializer.ex
```diff
@@ -145,6 +145,14 @@ defmodule JSONAPI.Serializer do
     merge_related_links(data, info, remove_links?())
   end
 
+  defp merge_base_links(doc, data, view, conn) do
+    links =
+      %{self: view.url_for(data, conn)}
```
doomspork: Could we avoid using a pipeline here?

https://github.com/christopheradams/elixir_style_guide#bare-variables

jherdman: Done!

doomspork: Thanks 😁

PR #166, opened 2019-01-27
Fix Bad Includes Returns HTTP 500 Response

### LINKED ISSUE #134: Bad include can return 500
According to the includes spec on jsonapi ([1](http://jsonapi.org/format/#fetching-includes)), invalid includes should only return 400.

https://github.com/jeregrine/jsonapi/blob/242499eaf42362fa0f1e6602ecc08f9baa7feaad/lib/jsonapi/query_parser.ex#L183

Looks to be due to this line. Based on whether or not the atom exists in the app, the line raises an error only on certain invalid includes

jherdman:
It looks like the smoking gun was in our tests this whole time.

Resolves #134

### DIFF
--- a/lib/jsonapi/plugs/query_parser.ex
+++ b/lib/jsonapi/plugs/query_parser.ex
@@ -182,22 +182,33 @@ defmodule JSONAPI.QueryParser do
       if inc =~ ~r/\w+\.\w+/ do
         acc ++ handle_nested_include(inc, valid_includes, config)
       else
-        inc = inc |> String.to_existing_atom()
+        inc =
+          try do
+            String.to_existing_atom(inc)
+          rescue
+            ArgumentError -> raise_invalid_include_query(inc, config.view.type())
+          end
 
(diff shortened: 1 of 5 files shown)


lib/jsonapi/plugs/query_parser.ex
```diff
@@ -182,7 +182,7 @@ defmodule JSONAPI.QueryParser do
       if inc =~ ~r/\w+\.\w+/ do
         acc ++ handle_nested_include(inc, valid_includes, config)
       else
-        inc = inc |> String.to_existing_atom()
+        inc = String.to_atom(inc)
```
jherdman: Are these changes a possible DOS vector?

jeregrine: Yea, it might be easier to try/catch around this?

jherdman: Done!

jherdman:
I've added a few type specs for the functions I touched. I'm not super familiar with writing them, so I'm open to either removing them entirely from this patch or feedback in improving them.

Issue #168, opened 2019-01-29
Add Typespecs for All of the Things


jherdman:
This ticket will be resolved when:

- [ ] Typespecs for all of the things
- [x] Dialyzer is run on CI, and is passing

jherdman:
I'm going to close this. The maintainers of this project are onboard and adding typespecs as they work.

PR #170, opened 2019-01-30
Dialyzer

jherdman:
Let's start resolving #168 by at least bringing Dialyzer into the CI pipeline.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,15 @@ env:
 notifications:
   email: false
 
+before_cache:
+  - rm -rf _build/$MIX_HOME/lib
+
+cache:
+  directories:
+    - _build/$MIX_HOME
+
 script:
   - mix test
   - mix credo
+  - mix dialyzer --halt-exit-status
   - if ! [[ `elixir -v` = *"1.5"* ]]; then mix format --check-formatted; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,5 +40,5 @@ cache:
 script:
   - mix test
   - mix credo
-  - mix dialyzer --halt-exit-status
+  - if [[ `elixir -v` = *"1.8"* ]]; then mix dialyzer --halt-exit-status; fi
   - if ! [[ `elixir -v` = *"1.5"* ]]; then mix format --check-formatted; fi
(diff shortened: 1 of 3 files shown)


doomspork:
@jherdman this is going to be awesome, nice work! Is caching the `_build` sufficient to capture the plts? I've had mixed success with the Guardian project's Travis config

jherdman:
@doomspork in looking over Guardian's Dialyxir config I think I've spotted the major difference:

https://github.com/ueberauth/guardian/blob/8930f7bfaaa3a382f0f1d2a87ed9337ee1f18dbb/mix.exs#L33

This option is deprecated according to the Dialyxir help (https://github.com/jeremyjh/dialyxir/blob/43f45e19c9735f12aa72cd3c328cf8a4dd9c513d/lib/mix/tasks/dialyzer.ex#L81). As such I tried to avoid using this option.

jherdman:
@doomspork in looking over Guardian's Dialyxir config I think I've spotted the major difference:

https://github.com/ueberauth/guardian/blob/8930f7bfaaa3a382f0f1d2a87ed9337ee1f18dbb/mix.exs#L33

This option is deprecated according to the Dialyxir help (https://github.com/jeremyjh/dialyxir/blob/43f45e19c9735f12aa72cd3c328cf8a4dd9c513d/lib/mix/tasks/dialyzer.ex#L81). As such I tried to avoid using this option.

jherdman:
TODO
------

```
lib/jsonapi/serializer.ex:56:pattern_match
The pattern
%{:type => _}
can never match the type
atom()
```

* Why did this only happen for one of our test runs?
* Should we constrain Dialyxir to run only for one one entry in our build matrix?

doomspork:
> Should we constrain Dialyxir to run only for one one entry in our build matrix?

Is this possible? I'm not a Travis guru.

jherdman:
@doomspork okay, so I clearly haven't added typespecs on all of the things, but this at least sets us off in the right direction. I'm only running Dialyzer on Elixir 1.8. This works around the weird error I noted above and leaves it unsolved on 1.7. Maybe this is okay, maybe it isn't. I propose that we merge this and deal with any fallout, and super duper promise to add typespecs for anything we touch going forward.

.travis.yml
```diff
@@ -30,7 +30,15 @@ env:
 notifications:
   email: false
 
+before_cache:
+  - rf -rf _build/$MIX_HOME/lib
```
doomspork: Do you mean `rm -rf`?

jherdman: I sure do... head -> desk

jherdman: Fixed!

PR #178, opened 2019-02-11
Globally Configurable Namespace

jherdman:
Why:

* A lot of the times a namespace is defined is to add the `/api` prefix
  to all the routes, and having to add it to all the views becomes tedious
  and error prone

* Allowing the namespace to be defined in a global config as a default
  that can be changed on a module by module basis

Note
-----

This pull request supersedes #130 by @zamith . Big ups to @zamith for the original effort 💪 .

### DIFF
--- a/lib/jsonapi/view.ex
+++ b/lib/jsonapi/view.ex
@@ -32,6 +32,10 @@ defmodule JSONAPI.View do
         end
       end
 
+      defmodule DogView do
+        use JSONAPI.View, namespace: "/pupperz-api"
+      end
+
   You can now call `UserView.show(user, conn, conn.params)` and it will render
   a valid jsonapi doc.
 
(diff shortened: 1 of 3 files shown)


lib/jsonapi/view.ex
```diff
@@ -104,6 +113,12 @@ defmodule JSONAPI.View do
         def type, do: raise("Need to implement type/0")
       end
 
+      if @namespace do
```
jherdman: This felt really dirty to me, though I have no real good reason to feel that way. Is there a better solution?

doomspork: I'm not entirely sure if there's an "ideal" way to handle this. We could skip the attribute and call `unquote(namespace)` directly?

```elixir
      if unquote(namespace) do
        def namespace, do: unquote(namespace)
      else
        def namespace, do: Application.get_env(:jsonapi, :namespace, "")
      end
```

Or maybe we just make it simple?


```elixir
def namespace, do: unquote(namespace) || Application.get_env(:jsonapi, :namespace, "")
```

jherdman: > Or maybe we just make it simple? ...

This version resulted in some weird warnings:

```elixir
warning: this check/guard will always yield the same result
  test/jsonapi/view_test.exs:5

warning: this check/guard will always yield the same result
  test/jsonapi/view_test.exs:19

warning: this check/guard will always yield the same result
  test/jsonapi/view_test.exs:43
```

These all point to the `use JSONAPI.View, namespace: "/foo"` invocations in our test views. I don't quite understand what's happening with that.

That said your first option worked, but didn't seem much better than what I had already. I've left my original solution in place. I'm more than open to changing it though.

doomspork: What you have is fine 😁

PR #179, opened 2019-02-11
Update Dialyzer CI Cache Steps

jherdman:
Dialyzer's README was updated to add common CI caching steps. I've
updated our procedure to match theirs.


doomspork:
Thank you @jherdman 😁

PR #180, opened 2019-02-12
Report on missing data[type] as such

kbaird:
Passing in a payload with `data: %{attributes: %{}}` would result in an error message stating that `attributes` was missing from `data`, when in fact what was missing was `type`. This corrects that.


jherdman:
@kbaird nice catch. Keep 'em comin'!

PR #185, opened 2019-03-09
Ensure Spec should set correct content-type header

### LINKED ISSUE #184: Setting content-type
When I render a view the content-type header is set to ``application/json`` but I'm using the Plug ``JSONAPI.EnsureSpec`` in the pipeline.
I think the problem is that phoenix is setting the header, but I don't know where. And then the Plug does not update the header which is the expect behavior. 
Also, I know it is working properly because when make an invalid request the header is set correctly to ``application/vnd.api+json``

0urobor0s:
This fixes #184

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,13 +122,15 @@ We include a set of Plugs to make enforcing the JSONAPI spec for requests easy. 
 plug JSONAPI.EnsureSpec
 ```
 
-Under-the-hood `JSONAPI.EnsureSpec` relies on three individual plugs:
+Under-the-hood `JSONAPI.EnsureSpec` relies on four individual plugs:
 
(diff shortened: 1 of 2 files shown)


jherdman:
Hi @0urobor0s! The README [discusses](https://github.com/jeregrine/jsonapi/tree/a1099c3af520eee9e8a1b4172db66776438b4e82#spec-enforcement) the `EnsureSpec` Plug. Would you mind updating this to include mention of your change?

doomspork:
These changes look good. Please update the README per @jherdman's feedback and we can merge 😁

jherdman:
Many thanks, @0urobor0s !

PR #131, opened 2018-09-12
Pagination with meta

zamith:
Refer to #104 for more details

PR #189, opened 2019-03-16
Reimplementing pagination

lucacorti:
This implements page/offset/cursor pagination strategies in the  serializer by passing a `JSONAPI.Page` object to the `serialize` function and implementing basic page, offset and cursor strategies. It also gets rid of the now useless `with_pagination` knob.


lucacorti:
After some thought, I think pagination strategies do not belong into the serializer. The JSONAPI spec does not mandate any particular strategy, it just reserves the use of the `page` query param for pagination.

So, my latest commit provides an extensible behavior for dealing with pagination which basically boils down to delegating pagination links to the view `pagination_links` overridable function.

The deafult implementation of the function provides no links, unless a `paginator` option is provided to `use JSONAPI.View` or a global `paginator` configuration is provided. These point to a module conforming to `JSONAPI.Paginator` which defines one callback function to delegate links generation to. This allows users to implement custom pagination strategies. The page struct from the conn assigns is used to retrieve total items/pages information which can be patched in the conn at the controller level.

An implementation of a page-based strategy is implemented. Tests are updated and demonstrate the usage.

I think this improves on the previous PR, thoughts?

doomspork:
Sorry for missing this PR @lucacorti! We'll get eyes on it this week and let you know our thoughts, thanks for getting involved 😁

lib/jsonapi/paginator.ex
```diff
@@ -0,0 +1,15 @@
...
+
+  @type links :: %{
+          first: String.t() | nil,
+          last: String.t() | nil,
+          next: String.t() | nil,
+          prev: String.t() | nil
+        }
+
+  @callback paginate(data :: term, view :: atom, conn :: Plug.Conn.t(), page :: JSONAPI.Page.t()) ::
+              links
```
doomspork: I don't think we want this on a newline, could you check that there's not extra whitespace or returns here?

lib/jsonapi/view.ex
```diff
@@ -140,6 +142,15 @@ defmodule JSONAPI.View do
         def namespace, do: Application.get_env(:jsonapi, :namespace, "")
       end
 
+      def pagination_links(data, conn, page) do
+        paginator = Application.get_env(:jsonapi, :paginator, @paginator)
+        if  Code.ensure_loaded?(paginator) && function_exported?(paginator, :paginate, 4) do
```
doomspork: Extra whitespace after the `if`

lib/jsonapi/paginator/page.ex
```diff
@@ -0,0 +1,30 @@
+defmodule JSONAPI.Paginator.Page do
+  @moduledoc """
+  Page based pagination strategy
+  """
+
+  @behaviour JSONAPI.Paginator
+
+  alias JSONAPI.Page
+
+  def paginate(data, view, conn, %Page{page: page, size: size, total_pages: total_pages}) do
```
doomspork: We need `@impl true` here.

lib/jsonapi/paginator/page.ex
```diff
@@ -0,0 +1,30 @@
...
+    first = view.url_for_pagination(data, conn, %{size: size, page: 1})
+    last = view.url_for_pagination(data, conn, %{size: size, page: total_pages})
+
+    next =
```
doomspork: I think with some pattern matching we can do away with these two `if` expressions:

```elixir
@impl true
  def paginate(data, view, conn, %{size: size, total_pages: total_pages} = page) do
    first = view.url_for_pagination(data, conn, %{size: size, page: 1})
    last = view.url_for_pagination(data, conn, %{size: size, page: total_pages})

    %{
      first: first, 
      last: last, 
      next: next_link(data, view, conn, page), 
      prev: previous_link(data, view, conn, page), 
    }
  end

  defp next_link(data, view, conn, %{page: page, size: size, total_pages: total_pages}) when page < total_pages,
    do: view.url_for_pagination(data, conn, %{size: size, page: page + 1})
  defp next_link(_data, _view, _conn, _page), 
    do: nil

  defp previous_link(data, view, conn, %{page: page, size: size}) when page > 1,
    do: view.url_for_pagination(data, conn, %{size: size, page: page - 1})
  defp previous_link(_data, _view, _conn, _page), 
    do: nil
```

If we ever need to change the logic for either of those it's already contained in it's own function.

lib/jsonapi/serializer.ex
```diff
@@ -146,29 +149,30 @@ defmodule JSONAPI.Serializer do
...
+  defp merge_base_links(%{links: links} = doc, data, view, conn) do
+    view_links =
+      data
+      |> view.links(conn)
+      |> Map.merge(links)
+      |> Map.merge(%{self: view.url_for(data, conn)})
 
-    Map.merge(doc, %{links: links})
+    doc
+    |> Map.merge(%{links: view_links})
```
doomspork: We try to avoid pipelines for single function calls. This should be expressed as: `Map.merge(doc, %{links: view_links})`

test/jsonapi/serializer_test.exs
```diff
@@ -536,39 +535,31 @@ defmodule JSONAPI.SerializerTest do
...
 
-  test "serialize includes pagination links if they are defined and with_pagination is configured" do
-    data = %{id: 1}
-    Application.put_env(:jsonapi, :with_pagination, true)
+  test "serialize includes pagination links if page-based pagination is requested" do
+    data = [%{id: 1}]
 
-    encoded = Serializer.serialize(PostView, data, nil)
+    conn =
+      %Plug.Conn{}
```
doomspork: Do we want to use `conn/3` here from Plug.Test?

test/jsonapi/serializer_test.exs
```diff
@@ -536,39 +535,31 @@ defmodule JSONAPI.SerializerTest do
...
+      |> Plug.Conn.assign(:jsonapi_query, %JSONAPI.Config{
```
doomspork: Let's not use a pipeline here since it's only a single function.

lib/jsonapi/paginator.ex
lucacorti: `mix fomat --check-formatted` fails if everything goes on the same line, not sure how to go about this one.

lib/jsonapi/view.ex
lucacorti: Fixed in latest commit, also catched by `mix format`

lib/jsonapi/paginator/page.ex
lucacorti: Fixed

lib/jsonapi/paginator/page.ex
lucacorti: Sure, fixed.

lib/jsonapi/serializer.ex
lucacorti: Noted, also fixed.

test/jsonapi/serializer_test.exs
lucacorti: I refactored the test to use Plug.Test.conn/3

test/jsonapi/serializer_test.exs
lucacorti: See below.

lib/jsonapi/paginator.ex
lucacorti: Aliased the modules to shorten line length seems like a goog compromise

lucacorti:
The failing test esposes a bug (#191) in the parser, which ends up populating the JSONAPI.Page struct with string values from the query parametes, which in turn break the paginator which expects integers as per `JSONAPI.Page` typespec.

jherdman:
Hey friends. Sorry for taking so long to get to this. Life, etc.

I haven't used the current pagination support, so I have some background questions:

- [ ] What are the problems with the current implementation?
- [ ] Does pagination even belong in the "core" of this project? Should it be extracted into a supporting addon?

lucacorti:
JSON API mandates the use of the `page` query params and defines standard pagination links format in the response, beyond that everything is implementation specific. So I think generic pagination support belongs in the library. My PR goes in this direction. The provided page-based implementation is merely an example implementation. I guess not shipping it or any other specific implementation would be fine, but could be nice to have out of the box.

The main problem with the current implementation from my point of view is everything is done into the links callback and doesn't offer much flexibility. In my opinion the links callback is a good place to address custom links, while being able to delegate pagination strategies allows to define a clear interface, configure pagination at project/view level and possibly override the pagination_links for even more flexibility.

So IMHO basic pagination support belongs in the project, specific pagination implementations maybe not.

Another thing I'd like to add to my PR to go down the spec compliance route even more is allowing the user to specify which query params map to which `JSONAPI.Page` struct members, because the usage of specific parameters is outside the scope of the spec and the library is enforcing the use of specific parameter names. Again, the current ones may stay as defaults, but the user should be able to verride them.

lucacorti:
Latest commit fixes #191 and adds support for custom query parameters for pagination via

```
config :jsonapi, page_query_param, "mypage"
```

For all pagination query params inside `page[]`.

jherdman:
Generally speaking I feel like this PR needs to be approached slightly differently. Currently:

* There's a fix for #191 in it
* It removes existing behaviour
* Adds new behaviour

I know this will be more work and it's a lot to ask from someone freely contributing their time, but I'd like to see an alternate approach.

1. Extract the fix for #191 into a distinct PR. This gives us the option of releasing it independently to see feedback from users of this package.
2. Construct this PR in a way that offers existing users of the pagination approach a cleaner, and documented, way to upgrade. My understanding is that this code would break existing consumption of this package, thus necessitating a major release without a release informing of a deprecated pathway, and a pathway to upgrade.

lucacorti:
@jherdman Sure, I'll make a separate PR for #191.

As far as this PR goes, documentation is the next step. If a major release is acceptable for the project, it should suffice, otherwise I'd like to discuss ways in which the upgrade path should be smoothed.

lucacorti:
@jherdman 

So, #192 addresses #191.

Also I gave some more thought about the upgrade path and realized my PR already addresses this in a backwards compatible way. Pagination links currently generated in the view `links` callback should be preserved if the user doesn't opt-in to the new pagination style by either implementing the `pagination_links` callback or by passing a paginator module to the view/config.

I guess properly documenting this behavior and how to migrate to the new one should be enough.

lucacorti:
So, I'd like to recap where I'm trying to go with this PR.

The current implementation has a few issues:

1. It does not expose a simple and clear interface to handle pagination.
2. It enforces a specific way to pass pagination parameters and the type of values int these which is not in line with the JSON API spec and limits pagination schemes that can be implemented.

What this PR does:

1. Moves pagination link generation to a new `pagination_links` callback in the view, which can be used to provide pagination links.
2. Provides a default implementation of the pagination links callback which delegates pagination to a module which handles pagination link generation through a well defined behavior. The pagination behavior can be implemented, configured globally, overriden per view.
3. Gets rid of the `JSONAPI.Page` struct and value parsing and just rely on the `page` query parameter contents to provide context for pagination, which is all JSON:API mandates. It also adds a way to pass needed context back to the pagination module (ie. total_items/total_pages) from the controller.
4. Provides backwards compatibility. If the user does not opt in to the new system, pagination links generated in the links callback should still work as expected.

What is missing to complete the work

1. Documentation about the new behaviour and how to opt in to the new pagination api.

lib/jsonapi/config.ex
```diff
@@ -3,16 +3,14 @@ defmodule JSONAPI.Config do
...
-
   defstruct data: nil,
             fields: %{},
             filter: [],
             include: [],
             opts: nil,
             sort: nil,
             view: nil,
-            page: %Page{}
+            page: %{}
```
jherdman: 100% agree that this is the right choice now that I understand this portion of the spec a little better.

lib/jsonapi/paginator.ex
```diff
@@ -0,0 +1,20 @@
+defmodule JSONAPI.Paginator do
```
jherdman: I think I understand what's going on here. Would I be correct in saying that this module should be implemented by the consumer of this lib to generate these links?

test/jsonapi/serializer_test.exs
```diff
@@ -536,39 +535,31 @@ defmodule JSONAPI.SerializerTest do
...
+      |> Plug.Test.conn("/mytype?page[page]=2&page[size]=1")
+      |> QueryParser.call(%Config{view: PaginatedPostView, opts: []})
 
-    assert encoded[:links][:next] ==
-             PostView.url_for_pagination(data, nil, %{cursor: "some-string"})
+    encoded =
+      Serializer.serialize(PaginatedPostView, data, conn, nil, total_pages: 3, total_items: 3)
 
-    Application.delete_env(:jsonapi, :with_pagination)
+    assert encoded[:links][:first]
```
jherdman: Should we consider actually asserting on what we expect the URL to look like?

jherdman:
How hard would it be to resolve the issues in #131 using your implementation?

lib/jsonapi/paginator.ex
lucacorti: A module conforming to the `Paginator` behavior can be either configured as the global pagination logic in `config :jsonapi, ;paginator, MODULE`, and/or passed to a view via `use JSONAPI.View, paginator: MODULE`. Its implementation can be done by the consumer of the library and/or by the library itself. I think shipping implementations for common pagination schemes (page/offset/cursor) with the library could be valueable, but I'd love to hear from you on this topic. This is also something that could be discussed separately from this PR.

test/jsonapi/serializer_test.exs
lucacorti: Sure, will update the tests.

lucacorti:
@jherdman I think this PR and #131 are solving the same problems in two fundamentally different ways. Which issues in #131 you are referring to?

lucacorti:
@jherdman About the `Paginator` behavior, its responsibility is to just generate links, actual pagination is done externally (in my case Scrivener paginating Ecto models), because knowledge about the actual data is needed to do that. No assumption is made about how pagination is actually performed.

jherdman:
> @jherdman I think this PR and #131 are solving the same problems in two fundamentally different ways. Which issues in #131 are you referring to?

Ah, my bad. I think we can ignore this comment.

lib/jsonapi/paginator.ex
jherdman: > This is also something that could be discussed separately from this PR.

Totally agreed. I'd rather see this come in to the lib first, have some people play with it and iron out basic bugs.

lucacorti: Ok, I'm going to move the page based implementation from the code to the docs as an example and we can figure out if including implementations with the library is feasible at a later time.

lucacorti:
@jherdman I've updated the tests and added documentation. I think this is quite complete. Anything else we need to address?

jherdman:
@lucacorti LGTM! It needs a rebase, but I'd love to get @doomspork or @snewcomer 's eyes on it too.

doomspork:
@jherdman / @lucacorti this looks good to me. Thank you both for putting in so much time to get this across the finish line.

lucacorti:
@jherdman  @doomspork Thank you for bearing with me while I iterated on the implementation.

doomspork:
Thank _you_ @lucacorti for doing that. It's truly _awesome_ to have open source contributors like you, I cannot stress enough how much we value people like yourself. I hope you'll continue your contributions 😁

PR #195, opened 2019-04-08
Change the key name for page numbers to number

strzibny:
According to the specification (https://jsonapi.org/format/#fetching-pagination) the page number should appear as `page[number]`, not `page[page]`. This PR fixes it.

### DIFF
--- a/lib/jsonapi/page.ex
+++ b/lib/jsonapi/page.ex
@@ -6,14 +6,14 @@ defmodule JSONAPI.Page do
   defstruct limit: nil,
             offset: nil,
             size: nil,
-            page: nil,
+            number: nil,
             cursor: nil
 
   @type t :: %__MODULE__{
           limit: nil | non_neg_integer,
           offset: nil | non_neg_integer,
           size: nil | non_neg_integer,
-          page: nil | non_neg_integer,
+          number: nil | non_neg_integer,
           cursor: nil | non_neg_integer
         }
 end
--- a/lib/jsonapi/plugs/query_parser.ex
+++ b/lib/jsonapi/plugs/query_parser.ex
@@ -48,7 +48,7 @@ defmodule JSONAPI.QueryParser do
         page: %JSONAPI.Page{
           limit: limit,
           offset: offset,
-          page: page,
+          number: number,
           size: size,
           cursor: cursor
         }}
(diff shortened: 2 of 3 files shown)


jeregrine:
This might be a breaking change.

jherdman:
Definitely. I'll think we'll have to support both with a deprecation pathway.

strzibny:
Yet, it is. As @jherdman mentions perhaps include both fields and issue warning somewhere?

lucacorti:
@strzibny That's just an example, no actual restriction is put on the contents of the `page` query param. I'd go with a more radical approach. Completely avoid parsing paging parameters into a struct.

strzibny:
>  no actual restriction is put on the contents of the page query param

yet the struct won't have this key, no?

> Completely avoid parsing paging parameters into a struct.

I like the fact it's parsed, there are not that many variants out there, right?

lucacorti:
@strzibny yes, but if the struct stays you should you should be able to configure which query parameters are mapped to which page struct fields. Also, this is tricky because the format of the parameters is also not mandated, so you can't assume they are integers or any other type.

This is why I did #192 but in retrospect I think the best way to address this would be to completely remove the `JSONAPI.Page` struct and directly use query params to handle this.

jherdman:
Hey friends. After having read over the spec I feel that we can close this issue. As @lucacorti pointed out, the confusion comes from this param being part of an example. The following are the _only_ keys in the pagination family that are prescribed:

* first: the first page of data
* last: the last page of data
* prev: the previous page of data
* next: the next page of data

Beyond that there are no prescriptions to the other parameters used.

Therefore I think we should close this ticket. I don't know if I'm ready to toss the Page struct yet, but clearly something needs to be done here.

strzibny:
@jherdman just to confirm, your recommendation is therefore be using `page[page]` for page number?

jherdman:
The JSON:API spec's [section on pagination](https://jsonapi.org/format/1.1/#fetching-pagination) makes no recommendation for key used for the current page number. The choice is yours!

strzibny:
My point is about the automatic assign:

```
conn.assigns.jsonapi_query.page.page
```
I won't have there "number". So if I want it I have to bypass `jsonapi_query` and do it on my own. Is that what you mean?

jherdman:
That's a great question and I honestly don't have a good answer for you at this time. If you find a work around please post it here. I'll try to free up some time to get the recent flurry of pagination work merged in and released.

lucacorti:
@strzibny  I'm working on fixing this in #189, but the current implementation forces you to use that specific query parameter name.

PR #197, opened 2019-04-10
Require ID for PATCH

kbaird:
Recently we've encountered an issue in which a PATCH payload that lacks an ID produces an incorrect error message stating that the `type` is what's missing.

This adds a new error that refers to `id`, and produces it in the cases we encountered.


jherdman:
Many thanks, @kbaird !

PR #198, opened 2019-04-19
Add some typespecs

kbaird:
Adds some typespecs

### DIFF
--- a/lib/jsonapi/error_view.ex
+++ b/lib/jsonapi/error_view.ex
@@ -72,8 +72,8 @@ defmodule JSONAPI.ErrorView do
     |> serialize_error
   end
 
-  @spec too_many_relationships_payload_for_standard_endpoint :: map()
-  def too_many_relationships_payload_for_standard_endpoint do
+  @spec to_many_relationships_payload_for_standard_endpoint :: map()
+  def to_many_relationships_payload_for_standard_endpoint do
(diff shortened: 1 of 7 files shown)


lib/jsonapi/error_view.ex
```diff
@@ -34,37 +37,43 @@ defmodule JSONAPI.ErrorView do
...
 
+  @spec missing_id :: map()
   def missing_id do
     "Missing id in data parameter"
     |> build_error(400, @crud_message, "/data/id")
     |> serialize_error
   end
 
-  def to_many_relationships_payload_for_standard_endpoint do
+  @spec too_many_relationships_payload_for_standard_endpoint :: map()
```
kbaird: I also fixed a typo here

doomspork: I wonder if we should continue to define "to_many" with a deprecation notice in the event folks are using these directly?

kbaird: My mistake. https://github.com/jeregrine/jsonapi/pull/198/commits/55dd0f2bf13a61b65302fabf1570318d5c492e3e reverts.

kbaird: My confusion sprang from the fact that both labels are appropriate. It is an error that pertains to a "to-many" relationship, and is triggered by their count being too high (i.e., there are "too many" of them).

PR #200, opened 2019-04-22
Serializer Includes Links by Default

### LINKED ISSUE #174: [RFC] Rework Relationships
## Summary

Relationship handling in the `View` layer is closely tied to the presence of data for a given relationship. This need not be the case when data isn't being sideloaded.

## Motivation

By removing the coupling between data presence and link inclusion we reduce overhead, and have a clearer API for relationship linking.

## Detailed Design

Let's introduce an overridable `relationships/1` method that is preferentially used when serializing to build the `"relationships"` object in a response. This method has the following signature:

```elixir
@spec relationships(conn :: Plug.Conn.t()) :: keyword()
```

In providing the `conn` argument we offer the ability for consumers of this library to prune, or add, relationships based upon, for instance, a current user.

In this method any and all configured relationships will have a link by default. Configured relationships will be sided loaded if, and only if, requested by the client (thus the `QueryParser` must be in play). A new option, `:embed`, will be introduced. This option is the same as `:include` — data will be embedded in the `"included"` object if, and only if, the associated data has been loaded.

In introducing this method we'll deprecate `relationships/0` with the intent of removing it in our next major release.

## Drawbacks

Additional `Serializer` and `View` complexity. We may need to do some refactoring here to reduce cognitive overhead whilst maintaining both pathways.

## Alternatives

- We could do nothing. The current implementation works.
- We willingly introduce a breaking change. The use case above wherein I talk about pruning a relationship for a user of lower privilege, for instance, might be better served with an entirely new view instead.

## Unresolved Questions

- [ ] I've reached out to @jeregrine regarding why we wanted the associated data to be loaded in the first place. This decision goes quite a ways back: b8bc80c5549bb7bb07934da762d888504365f90e. There may be really good reasons for this that I'm missing.
- [x] Should `data` be provided to the new `relationships` method too? I can't really see why it would be needed, and I feel like it's best to constrain whatever we add as much as possible. One advantage of doing so would be that edge cases like #136 could be resolved (though I'm not entirely sure how this would fit in with the JSON:API spec).

jherdman:
Relationships will always, at the least, include a `"self"` link unless
`remove_links` has been configured.

This aligns us more closely with the spec. According to the spec a
relationship object MUST contain one of:

* `links`
* `data`
* `meta`

Before this change it was possible to generate an empty relationship
object. The cost of this is increased payload size, though I think most
users will find this OK.

Resolves #174

### DIFF
--- a/lib/jsonapi/serializer.ex
+++ b/lib/jsonapi/serializer.ex
@@ -76,9 +76,8 @@ defmodule JSONAPI.Serializer do
   end
 
   @spec encode_relationships(Plug.Conn.t(), serialized_doc(), tuple(), list()) :: tuple()
-  def encode_relationships(conn, doc, {view, data, _, _} = view_info, options) do
+  def encode_relationships(conn, doc, {view, _data, _, _} = view_info, options) do
     view.relationships()
-    |> Enum.filter(&data_loaded?(Map.get(data, elem(&1, 0))))
     |> Enum.map_reduce(doc, &build_relationships(conn, view_info, &1, &2, options))
   end
 
@@ -152,11 +151,15 @@ defmodule JSONAPI.Serializer do
 
   @spec encode_relation(tuple()) :: map()
   def encode_relation({rel_view, rel_data, _rel_url, _conn} = info) do
-    data = %{
-      data: encode_rel_data(rel_view, rel_data)
-    }
+    if data_loaded?(rel_data) do
+      data = %{
+        data: encode_rel_data(rel_view, rel_data)
+      }
 
-    merge_related_links(data, info, remove_links?())
+      merge_related_links(data, info, remove_links?())
+    else
+      merge_related_links(%{}, info, remove_links?())
+    end
   end
 
   defp merge_base_links(%{links: links} = doc, data, view, conn) do
@@ -188,7 +191,12 @@ defmodule JSONAPI.Serializer do
          {rel_view, rel_data, rel_url, conn},
          false = _remove_links
        ) do
-    Map.merge(encoded_data, %{links: %{self: rel_url, related: rel_view.url_for(rel_data, conn)}})
+    if data_loaded?(rel_data) do
+      related_url = rel_view.url_for(rel_data, conn)
+      Map.merge(encoded_data, %{links: %{self: rel_url, related: related_url}})
+    else
+      Map.merge(encoded_data, %{links: %{self: rel_url}})
+    end
   end
 
   defp merge_related_links(encoded_rel_data, _info, _remove_links), do: encoded_rel_data
--- a/mix.exs
+++ b/mix.exs
@@ -53,6 +53,7 @@ defmodule JSONAPI.Mixfile do
       {:earmark, ">= 0.0.0", only: :dev},
       {:credo, "~> 1.0", only: [:dev, :test], runtime: false},
       {:phoenix, "~> 1.3", only: :test},
+      {:ecto, "~> 3.1", only: :test},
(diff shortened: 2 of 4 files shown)


lib/jsonapi/serializer.ex
```diff
@@ -76,9 +76,8 @@ defmodule JSONAPI.Serializer do
   end
 
   @spec encode_relationships(Plug.Conn.t(), serialized_doc(), tuple(), list()) :: tuple()
-  def encode_relationships(conn, doc, {view, data, _, _} = view_info, options) do
+  def encode_relationships(conn, doc, {view, _data, _, _} = view_info, options) do
     view.relationships()
-    |> Enum.filter(&data_loaded?(Map.get(data, elem(&1, 0))))
```
jherdman: This is the source of the breaking change. I think we could make the new behaviour opt-in at the cost of greater complexity.

lib/jsonapi/serializer.ex
```diff
@@ -188,7 +191,12 @@ defmodule JSONAPI.Serializer do
          {rel_view, rel_data, rel_url, conn},
          false = _remove_links
        ) do
-    Map.merge(encoded_data, %{links: %{self: rel_url, related: rel_view.url_for(rel_data, conn)}})
+    if rel_data do
+      related_url = rel_view.url_for(rel_data, conn)
+      Map.merge(encoded_data, %{links: %{self: rel_url, related: related_url}})
+    else
+      Map.merge(encoded_data, %{links: %{self: rel_url}})
```
jherdman: We might be able to generate the `related` URL if there's a foreign key in the data

jherdman:
Smoke testing in my app shows that this isn't working quite like I expected. I'll have to add some more tests to show the edge cases I'm trying to avoid.

jherdman:
OK. I'm reasonably sure this works now and could use some feedback from some smart folks in the community.

@strzibny how well does this patch resolve your needs from #174?

lib/jsonapi/serializer.ex
snewcomer: So the change here is if no data, then the empty relationship is included?  Is there a test that shows the change?

strzibny:
One thing this does not address is that it always include a relationship *self* link. I would like to build API completely without providing this addresses. Now they are included, but not implemented by me. I only want to include *related* link and that's it. `ja_serializer` for instance allow to configure this links for relationships and it's easy to omit it.

jherdman:
> One thing this does not address is that it always include a relationship self link. I would like to build API completely without providing this addresses. Now they are included, but not implemented by me. I only want to include related link and that's it

Argh.... Yeah, I have to admit I was hoping to work around this for a bit as I wrap my head around the full extent of the problem.

> ... does jsonapi provide the views for this relationship URLs?

We provide a few [overridable methods](https://github.com/jeregrine/jsonapi/blob/fcaceabe72aa2d407df6a241552afdf61910b464/lib/jsonapi/view.ex#L214-L239) to set up links. I'm not entirely certain if we support relationship `meta` (I think we do).

OK. I'll take another kick at this. I'm fairly convinced we'll have to introduce a breaking change to the library, but maybe that's alright?

strzibny:
Personally I am not against breaking changes *in this case*. In my mind this is still young project that needs a little bit more flexibility first. I am also thinking if I should just give up and implement the *self* relationship links, but then again - spec don't require them and sometimes you just can live without them IMHO.

jherdman:
This is true. The spec only says that `links` must have one of `self` or `related` — not necessarily both. I wonder if we can provide some sort of tooling generate the `self` links? They're kind of a pain.

jherdman:
Closing. I'll take another kick at this soon.

PR #203, opened 2019-05-10
Fix JSONApi sparse fieldset compliance for relationships

leepfrog:
Changes:
- query_parser: change fields resolution to be based on type instead of relationship name.
  see: https://jsonapi.org/format/#fetching-sparse-fieldsets
- query_parser_test: fix a typo in test relationship definition

Relevant section from the spec:  "A client MAY request that an endpoint return only specific fields in the response on a per-type basis by including a fields[TYPE] parameter."

### DIFF
--- a/lib/jsonapi/plugs/query_parser.ex
+++ b/lib/jsonapi/plugs/query_parser.ex
@@ -246,7 +246,7 @@ defmodule JSONAPI.QueryParser do
 
   @spec get_view_for_type(module(), String.t()) :: module() | no_return()
   def get_view_for_type(view, type) do
-    case Enum.find(view.relationships, fn {k, _v} -> Atom.to_string(k) == type end) do
+    case Enum.find(view.relationships, fn {_k, v} -> v.type == type end) do
       {_, view} -> view
       nil -> raise_invalid_field_names(type, view.type())
     end
--- a/lib/jsonapi/plugs/query_parser.ex
+++ b/lib/jsonapi/plugs/query_parser.ex
@@ -1,6 +1,6 @@
 defmodule JSONAPI.QueryParser do
   @behaviour Plug
-  alias JSONAPI.Config
+  alias JSONAPI.{Config, Deprecation}
   alias JSONAPI.Exceptions.InvalidQuery
   alias Plug.Conn
   import JSONAPI.Utils.IncludeTree
@@ -246,9 +246,26 @@ defmodule JSONAPI.QueryParser do
 
   @spec get_view_for_type(module(), String.t()) :: module() | no_return()
   def get_view_for_type(view, type) do
-    case Enum.find(view.relationships, fn {_k, v} -> v.type == type end) do
+    case Enum.find(view.relationships, fn relationship ->
+           is_field_valid_for_relationship(relationship, type)
+         end) do
       {_, view} -> view
       nil -> raise_invalid_field_names(type, view.type())
+    end
+  end
+
+  @spec is_field_valid_for_relationship({atom(), module()}, String.t()) :: boolean()
+  defp is_field_valid_for_relationship({key, view}, expected_type) do
+    cond do
+      view.type == expected_type ->
+        true
+
+      Atom.to_string(key) == expected_type ->
+        Deprecation.warn(:query_parser_fields)
+        true
+
+      true ->
+        false
     end
   end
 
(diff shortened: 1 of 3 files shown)


test/jsonapi/plugs/query_parser_test.exs
```diff
@@ -14,7 +14,7 @@ defmodule JSONAPI.QueryParserTest do
       [
         author: JSONAPI.QueryParserTest.UserView,
         comments: JSONAPI.QueryParserTest.CommentView,
-        best_friends: JSONAPI.QueryParsertTest.UserView
+        best_friends: JSONAPI.QueryParserTest.UserView
```
jherdman: Nice catch! How is this not causing any issues in our test suite?

lib/jsonapi/plugs/query_parser.ex
```diff
@@ -246,7 +246,7 @@ defmodule JSONAPI.QueryParser do
 
   @spec get_view_for_type(module(), String.t()) :: module() | no_return()
   def get_view_for_type(view, type) do
-    case Enum.find(view.relationships, fn {k, _v} -> Atom.to_string(k) == type end) do
+    case Enum.find(view.relationships, fn {_k, v} -> v.type == type end) do
```
jherdman: It looks to me that although we're matching the spec more closely with this change that it _could_ break existing user's apps. I think we'll need to support looking for a view on both the `type` and key for now, and log a deprecation warning that we'll only support the `type` parameter in 2.0.

We don't have a formal deprecation logging mechanism in play right now, but in the past we've done this:

https://github.com/jeregrine/jsonapi/commit/323eab356ff309e7d383c63d7f612d662771fd72#diff-145509c96e7086ca2b927dd4897ed875L10

test/jsonapi/plugs/query_parser_test.exs
```diff
@@ -140,8 +140,8 @@ defmodule JSONAPI.QueryParserTest do
   end
 
   test "get_view_for_type/2 raises on invalid fields" do
-    assert_raise InvalidQuery, "invalid fields, comment for type mytype", fn ->
-      get_view_for_type(MyView, "comment")
+    assert_raise InvalidQuery, "invalid fields, comments for type mytype", fn ->
```
jherdman: Per the above deprecation pathway, this test should be noted as the deprecated pathway and an additional test added to show the sanctioned way forward.

test/jsonapi/plugs/query_parser_test.exs
leepfrog: I don’t believe the code is interpreted until accessed?  (It started to complain / fail the test suite once I fixed the underlying bug)

leepfrog:
Thx!  Will make these changes and update this PR shortly.

test/jsonapi/plugs/query_parser_test.exs
jherdman: Module names are just atoms, right? So that means you can just type any module name and Elixir is okay. I found this to be a source of frustration when learning Elixir. Here's a surprising example from an IEx session:

```
~ % iex
Erlang/OTP 21 [erts-10.3.4] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [hipe] [dtrace]

Interactive Elixir (1.8.1) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)> Foo.Bar.Baz
Foo.Bar.Baz
```

Boom. Totally okay.

leepfrog: That explains it!  Thanks!

lib/jsonapi/plugs/query_parser.ex
leepfrog: Added deprecation warning and restored original behavior (to run second after the preferred behavior)

test/jsonapi/plugs/query_parser_test.exs
leepfrog: Added a third test to capture the three cases and label on the deprecated test case.

lib/jsonapi/plugs/query_parser.ex
```diff
@@ -246,12 +246,29 @@ defmodule JSONAPI.QueryParser do
...
-    case Enum.find(view.relationships, fn {k, _v} -> Atom.to_string(k) == type end) do
+    case Enum.find(view.relationships, fn relationship ->
+           is_field_valid_for_relationship(relationship, type)
+         end) do
       {_, view} -> view
       nil -> raise_invalid_field_names(type, view.type())
     end
   end
 
+  @spec is_field_valid_for_relationship({atom(), module()}, String.t()) :: boolean()
```
leepfrog: This was the simplest way I could think of to handle the deprecation, but happy to take suggestions if needed.

PR #206, opened 2019-06-04
Use string status codes in error responses

jasondew:
The [JSON:API spec for errors](https://jsonapi.org/format/#errors) says that the status should be "expressed as a string value" so this PR makes that change!


jherdman:
Nice catch!

PR #208, opened 2019-06-17
Fix #207 problems with 'self' pagination link

### LINKED ISSUE #207: Problems with `self` pagination link
Hi. We're trying to upgrade to a version that includes https://github.com/jeregrine/jsonapi/pull/189, but have run into some problems.

Here's an example of some test regressions we have:
```elixir
  9) test index respects the page number 7 and size 3 (OurAppWeb.FooControllerTest)
     test/our_app_web/controllers/foo_controller_test.exs:342
     Assertion with == failed
     code:  assert resp["links"] == expected_pagination_links("foos", page_params)
     left:  %{"first" => "/api/v1/foos?page[page]=1&page[page-size]=3", "last" => "/api/v1/foos?page[page]=10&page[page-size]=3", "next" => "/api/v1/foos?page[page]=8&page[page-size]=3", "prev" => "/api/v1/foos?page[page]=6&page[page-size]=3", "self" => "http://localhost/foos"}
     right: %{"first" => "/api/v1/foos?page[page]=1&page[page-size]=3", "last" => "/api/v1/foos?page[page]=10&page[page-size]=3", "next" => "/api/v1/foos?page[page]=8&page[page-size]=3", "prev" => "/api/v1/foos?page[page]=6&page[page-size]=3", "self" => "/api/v1/foos?page[page]=7&page[page-size]=3"}
     stacktrace:
       test/our_app_web/controllers/foo_controller_test.exs:361: (test)
```

You see that the `self` value lacks the pagination params, pointing instead to a raw top-level `/foos` endpoint. I presume this comes from [view.url_for/2 here](https://github.com/jeregrine/jsonapi/pull/189/files?file-filters%5B%5D=.ex&file-filters%5B%5D=.exs&hide-deleted-files=true#diff-c6eda26dd75d06ea8b08bf45cb6f3f33R167)

In addition, I notice that [`@type links` in `paginator.ex` lacks the `self` key](https://github.com/jeregrine/jsonapi/pull/189/files?file-filters%5B%5D=.ex&file-filters%5B%5D=.exs&hide-deleted-files=true#diff-0e2f0db679e38d3135ee38439ff91b52R12). This seems at least tangentially related.

thebrianemory:
I was working with @kbaird when we came across the issue with #207.  This combines the version of `merge_base_links/4` previous to #189 and the implementation it introduced. This keeps all tests green while fixing our issue with `self` pagination links.

### DIFF
--- a/lib/jsonapi/serializer.ex
+++ b/lib/jsonapi/serializer.ex
@@ -161,7 +161,9 @@ defmodule JSONAPI.Serializer do
 
   defp merge_base_links(%{links: links} = doc, data, view, conn) do
     view_links =
-      Map.merge(%{self: view.url_for(data, conn)}, view.links(data, conn)) |> Map.merge(links)
+      %{self: view.url_for(data, conn)}
+      |> Map.merge(view.links(data, conn))
+      |> Map.merge(links)
 
     Map.merge(doc, %{links: view_links})
   end
--- a/lib/jsonapi/serializer.ex
+++ b/lib/jsonapi/serializer.ex
@@ -161,10 +161,7 @@ defmodule JSONAPI.Serializer do
 
   defp merge_base_links(%{links: links} = doc, data, view, conn) do
     view_links =
-      data
-      |> view.links(conn)
-      |> Map.merge(links)
-      |> Map.merge(%{self: view.url_for(data, conn)})
+      Map.merge(%{self: view.url_for(data, conn)}, view.links(data, conn)) |> Map.merge(links)
 
     Map.merge(doc, %{links: view_links})
   end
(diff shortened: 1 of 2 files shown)


snewcomer:
Looks good!  If you run `mix credo --strict`, there is a small error that hopefully should make the tests pass!

thebrianemory:
> Is it possible to add a test that would have caught this regression?

Sure thing!

thebrianemory:
Updated `SerializerTest` to return a proper self link. This included adding a self link for the test `Paginator` and updating the test on line 585:
"serialize includes pagination links if page-based pagination is requested".

It now asserts that the result of calling `Serializer.serialize/5` will return a properly formatted
self link. This can be verified by changing how `view_links` is created in
`Serializer.merge_base_links/4`.

The above mentioned test will pass with the changes in my PR
```elixir
view_links =
  %{self: view.url_for(data, conn)}
  |> Map.merge(view.links(data, conn))
  |> Map.merge(links)
```
but will fail with the previous version
```elixir
view_links =
  data
  |> view.links(conn)
  |> Map.merge(links)
  |> Map.merge(%{self: view.url_for(data, conn)})
```

This should help against future regressions.

PR #211, opened 2019-09-27
Fix `UnderscoreParameters` with Camelized Keys

jherdman:


lib/jsonapi/utils/string.ex
```diff
@@ -24,10 +24,16 @@ defmodule JSONAPI.Utils.String do
       iex> underscore("-top--posts-")
       "-top--posts-"
 
+      iex> underscore("corgiAge")
+      "corgi_age"
```
snewcomer: Purely curious - what cases are you finding for underscoring a camelized value?  Is dasherizing a camelized value also an applicable use case?

jherdman: My API accepts parameters that are camelized, e.g. a POST might look like:

```json
{
  "data": {
    "attributes": {
      "firstName": James
// snip
```

The `UnderscoreParameters` module wasn't transforming these into underscore values for ease of use with Elixir (which is how I caught this to begin with).

snewcomer: Ah ok 👍 Just incoming fields...

PR #214, opened 2019-10-23
Add query parameters to the links

komlanvi:
Query parameters are missing from the generated payload and pagination links which only include pagination information (page offset and size).

### DIFF
--- a/lib/jsonapi/utils/list.ex
+++ b/lib/jsonapi/utils/list.ex
@@ -4,10 +4,6 @@ defmodule JSONAPI.Utils.List do
   @doc """
   Transforms a Map into a List of Tuples that can be converted into a query string via URI.encode_query/1
 
-  - if values are terms that implements the String.Chars protocol it returns [{key1, value1}, {key2, value2}]
-  - if any value is a list es. %{key1: ["a", "b"]} it returns [{key1[], "a"}, {key1[], "b"}]
-  - if any value is a map es. %{key1: %{key2: "c", key3: "d"}} it returns [{key1[key2], "c"}, {key1[key3], "d"}]
-
   ## Examples
 
       iex> to_list_of_query_string_components(%{"number" => 5})
--- a/lib/jsonapi/utils/list.ex
+++ b/lib/jsonapi/utils/list.ex
@@ -1,0 +1,50 @@
+defmodule JSONAPI.Utils.List do
+  @moduledoc """
+  Utility functions to build a list from a map in a different format than the one returned by Map.to_list/2
+  """
+
+  @doc """
+  Transform a map es. %{key1: value1, key2: value2} into a list of two element tuple.
+
+  - if values are terms that implements the String.Chars protocol it returns [{key1, value1}, {key2, value2}]
+  - if any value is a list es. %{key1: ["a", "b"]} it returns [{key1[], "a"}, {key1[], "b"}]
+  - if any value is a map es. %{key1: %{key2: "c", key3: "d"}} it returns [{key1[key2], "c"}, {key1[key3], "d"}]
+
+  ## Examples
+
+      iex> to_custom_list(%{"number" => 5})
+      [{"number", 5}]
+
+      iex> to_custom_list(%{color: "red"})
+      [{:color, "red"}]
+
+      iex> to_custom_list(%{"alphabet" => ["a", "b", "c"]})
+      [{"alphabet[]", "a"}, {"alphabet[]", "b"}, {"alphabet[]", "c"}]
+
+      iex> to_custom_list(%{"filters" => %{"age" => 18, "name" => "John"}})
+      [{"filters[age]", 18}, {"filters[name]", "John"}]
+
+  """
+  @spec to_custom_list(map()) :: list(tuple())
+  def to_custom_list(map) when is_map(map) do
+    Enum.flat_map(map, &do_to_custom_list/1)
+  end
+
+  defp do_to_custom_list({key, value}) when is_list(value) do
+    to_list_of_tuple(key, value)
+  end
+
+  defp do_to_custom_list({key, value}) when is_map(value) do
+    Enum.flat_map(value, fn {k, v} -> to_list_of_tuple("#{key}[#{k}]", v) end)
+  end
+
+  defp do_to_custom_list({key, value}), do: to_list_of_tuple(key, value)
+
+  defp to_list_of_tuple(key, value) when is_list(value) do
+    Enum.map(value, &{"#{key}[]", &1})
+  end
+
+  defp to_list_of_tuple(key, value) do
+    [{key, value}]
+  end
+end
--- a/lib/jsonapi/utils/list.ex
+++ b/lib/jsonapi/utils/list.ex
@@ -29,20 +29,20 @@ defmodule JSONAPI.Utils.List do
   end
 
   defp do_to_list_of_query_string_components({key, value}) when is_list(value) do
-    into_tuple_and_put_in_list(key, value)
+    to_list_of_two_elem_tuple(key, value)
   end
 
   defp do_to_list_of_query_string_components({key, value}) when is_map(value) do
-    Enum.flat_map(value, fn {k, v} -> into_tuple_and_put_in_list("#{key}[#{k}]", v) end)
+    Enum.flat_map(value, fn {k, v} -> to_list_of_two_elem_tuple("#{key}[#{k}]", v) end)
   end
 
-  defp do_to_list_of_query_string_components({key, value}), do: into_tuple_and_put_in_list(key, value)
+  defp do_to_list_of_query_string_components({key, value}), do: to_list_of_two_elem_tuple(key, value)
 
-  defp into_tuple_and_put_in_list(key, value) when is_list(value) do
+  defp to_list_of_two_elem_tuple(key, value) when is_list(value) do
     Enum.map(value, &{"#{key}[]", &1})
   end
 
-  defp into_tuple_and_put_in_list(key, value) do
+  defp to_list_of_two_elem_tuple(key, value) do
     [{key, value}]
   end
 end
--- a/lib/jsonapi/utils/list.ex
+++ b/lib/jsonapi/utils/list.ex
@@ -1,10 +1,8 @@
 defmodule JSONAPI.Utils.List do
-  @moduledoc """
-  Utility functions to build a list from a map in a different format than the one returned by Map.to_list/2
-  """
+  @moduledoc false
 
   @doc """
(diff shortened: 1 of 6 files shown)


lib/jsonapi/view.ex
```diff
@@ -229,15 +229,21 @@ defmodule JSONAPI.View do
...
       end
 
-      def url_for_pagination(data, conn, pagination_attrs) do
-        pagination_attrs
-        |> Enum.reduce(%{}, fn {key, value}, acc ->
-          Map.put(acc, "page[#{key}]", value)
-        end)
+      def url_for_pagination(data, %{query_params: query_params} = conn, pagination_attrs) do
+        query_params
+        |> Map.drop([:__struct__])
```
doomspork: Why do we need this `Map.drop([:__struct__])`? Looking at the `Plug.Conn` documentation, the `query_params` are just a map, not a struct.

lib/jsonapi/view.ex
```diff
@@ -217,7 +217,7 @@ defmodule JSONAPI.View do
 
       def url_for(data, nil), do: "#{namespace()}/#{type()}/#{id(data)}"
 
-      def url_for(data, %Plug.Conn{} = conn) when is_list(data) do
+      def url_for(data, %Plug.Conn{} = conn) when is_list(data) or is_nil(data) do
```
doomspork: Why is this new guard necessary?

test/jsonapi/view_test.exs
```diff
@@ -146,10 +146,17 @@ defmodule JSONAPI.ViewTest do
...
+    conn = Plug.Conn.fetch_query_params(%Plug.Conn{})
 
-    assert PostView.url_for_pagination(nil, nil, %{number: 1, size: 10}) ==
-             "/api/posts?page%5Bnumber%5D=1&page%5Bsize%5D=10"
+    assert PostView.url_for_pagination(nil, conn, %{}) == "http://www.example.com/api/posts"
+
+    assert PostView.url_for_pagination(nil, conn, %{number: 1, size: 10}) ==
+             "http://www.example.com/api/posts?page%5Bnumber%5D=1&page%5Bsize%5D=10"
+
+    conn_with_query_params = Kernel.update_in(conn.query_params, &Map.put(&1, :number, 5))
```
doomspork: Small detail: per the docs `:number` is expected to be `"number"`.

test/jsonapi/view_test.exs
```diff
@@ -146,10 +146,17 @@ defmodule JSONAPI.ViewTest do
   end
 
   test "url_for_pagination/3" do
-    assert PostView.url_for_pagination(nil, nil, %{}) == "/api/posts"
+    conn = Plug.Conn.fetch_query_params(%Plug.Conn{})
```
doomspork: What do you think about pulling these changes out into their own tests, specifically testing the retention of query params on pagination links?

lib/jsonapi/view.ex
```diff
@@ -229,15 +229,21 @@ defmodule JSONAPI.View do
...
+      def url_for_pagination(data, %{query_params: query_params} = conn, pagination_attrs) do
+        query_params
+        |> Map.drop([:__struct__])
+        |> Map.put("page", pagination_attrs)
+        |> Enum.flat_map(&build_query_params/1)
         |> URI.encode_query()
         |> prepare_url(data, conn)
       end
 
+      defp build_query_params({key, %{} = value}) do
```
doomspork: Will values always be maps?

lib/jsonapi/view.ex
```diff
@@ -229,15 +229,21 @@ defmodule JSONAPI.View do
...
+        Enum.map(value, fn {k, v} -> {"#{key}[#{k}]", v} end)
+      end
+
+      defp build_query_params(data), do: [data]
```
doomspork: For my own edification, when will this be matched and used?

doomspork:
We also need to make sure the build passes @komlanvi 👍

lib/jsonapi/view.ex
komlanvi: Without it, this test fail
```
1) test serialize includes pagination links if page-based pagination is requested (JSONAPI.SerializerTest)
     test/jsonapi/serializer_test.exs:585
     ** (Protocol.UndefinedError) protocol Enumerable not implemented for %{:__struct__ => Plug.Conn.Unfetched, :aspect => :query_params, "page" => %{"page" => "1", "size" => "1"}}. This protocol is implemented for: GenEvent.Stream, Range, HashDict, Function, HashSet, MapSet, Date.Range, Stream, Map, IO.Stream, List, File.Stream
     code: Serializer.serialize(PaginatedPostView, data, conn, nil, total_pages: 3, total_items: 3)
     stacktrace:
       (elixir) /home/build/elixir/lib/elixir/lib/enum.ex:1: Enumerable.impl_for!/1
       (elixir) /home/build/elixir/lib/elixir/lib/enum.ex:141: Enumerable.reduce/3
       (elixir) lib/enum.ex:3015: Enum.flat_map/2
       test/jsonapi/serializer_test.exs:69: JSONAPI.SerializerTest.PaginatedPostView.url_for_pagination/3
       test/jsonapi/serializer_test.exs:45: JSONAPI.SerializerTest.PageBasedPaginator.paginate/5
       (jsonapi) lib/jsonapi/serializer.ex:179: JSONAPI.Serializer.merge_links/7
       test/jsonapi/serializer_test.exs:595: (test)
```
But I only have to call `Plug.Conn.fetch_query_params` on the `conn` before passing it through the test instead of the solution above.

lib/jsonapi/view.ex
komlanvi: Only map and list should be pattern matched. Others accepted value type in the url should be directly converted to string.

lib/jsonapi/view.ex
```diff
@@ -229,15 +229,32 @@ defmodule JSONAPI.View do
...
+        Enum.flat_map(value, fn {k, v} -> format("#{key}[#{k}]", v) end)
+      end
+
+      defp build_query_params({key, value}), do: format(key, value)
+
+      defp format(key, value) when is_list(value) do
+        Enum.map(value, & {"#{key}[]", &1})
+      end
+
+      defp format(key, value) do
```
jherdman: IMHO this function name is a bit too generic. "`format` what?"

lib/jsonapi/view.ex
komlanvi: or add possibility to return `%{data: null, etc...}` if render was called with `data = nil` . This must be the case specified in https://jsonapi.org/format/#document-top-level in the phrase 
```
Primary data MUST be either:

 - a single resource object, a single resource identifier object, or null, for requests that target single resources
- an array of resource objects, an array of resource identifier objects, or an empty array ([]), for requests that target resource collections
```

komlanvi:
Hi @doomspork @jherdman any news about this?

lib/jsonapi/utils/list.ex
```diff
@@ -0,0 +1,50 @@
...
+
+      iex> to_custom_list(%{"alphabet" => ["a", "b", "c"]})
+      [{"alphabet[]", "a"}, {"alphabet[]", "b"}, {"alphabet[]", "c"}]
+
+      iex> to_custom_list(%{"filters" => %{"age" => 18, "name" => "John"}})
+      [{"filters[age]", 18}, {"filters[name]", "John"}]
+
+  """
+  @spec to_custom_list(map()) :: list(tuple())
+  def to_custom_list(map) when is_map(map) do
```
jherdman: I'm not super jazzed with this method name. What does "custom list" mean? Can I (the person using this lib) "customize" the list?

lib/jsonapi/utils/list.ex
```diff
@@ -0,0 +1,50 @@
+defmodule JSONAPI.Utils.List do
+  @moduledoc """
+  Utility functions to build a list from a map in a different format than the one returned by Map.to_list/2
+  """
+
+  @doc """
+  Transform a map es. %{key1: value1, key2: value2} into a list of two element tuple.
```
jherdman: "es"?

lib/jsonapi/utils/list.ex
komlanvi: Hi thank you for your feedback. What about `to_list_of_two_elem_tuple/1` ?

lib/jsonapi/utils/list.ex
```diff
@@ -0,0 +1,50 @@
...
+
+  - if values are terms that implements the String.Chars protocol it returns [{key1, value1}, {key2, value2}]
+  - if any value is a list es. %{key1: ["a", "b"]} it returns [{key1[], "a"}, {key1[], "b"}]
+  - if any value is a map es. %{key1: %{key2: "c", key3: "d"}} it returns [{key1[key2], "c"}, {key1[key3], "d"}]
```
jherdman: It looks like these should be promoted to the examples below

lib/jsonapi/utils/list.ex
```diff
@@ -0,0 +1,50 @@
+defmodule JSONAPI.Utils.List do
+  @moduledoc """
```
jherdman: IMHO `@moduledoc false` -- we don't want people using this.

lib/jsonapi/utils/list.ex
jherdman: Hmm... I guess that's what it literally does, but it doesn't describe why the transformations you implemented take place. How about `to_list_of_query_string_components`?

lib/jsonapi/utils/list.ex
jherdman: "Transforms a Map into a List of Tuples that can be converted into a query string via URI.encode_query/1"

komlanvi:
Any comment ? @doomspork @jeregrine @mitchellhenke @snewcomer

lib/jsonapi/utils/list.ex
```diff
@@ -0,0 +1,44 @@
+defmodule JSONAPI.Utils.List do
+  @moduledoc false
+
+  @doc """
+  Transforms a Map into a List of Tuples that can be converted into a query string via URI.encode_query/1
+
+  ## Examples
```
jherdman: 💪

doomspork:
Things look good to me :grin:

doomspork:
Thank you @komlanvi :tada:

lucacorti:
Hi,

any plan for a `1.2.3` release to address links with query params?

thanks

jherdman:
Hi Luca! My apologies, I'll try to get this cut today.

jherdman:
@lucacorti done!

lucacorti:
@jherdman thanks a bunch

Issue #217, opened 2020-01-06
Serialize a List of data structures. 


ckoch-cars:
```defmodule MyApp.ExampleView do
  use JSONAPI.View, type: "exampleView"
  def fields, do: [:a_one, :b_two, :c_three]
  ...

  # specify the functions to destructure the Structs into Maps.
  def a_one(%_module{} = struct), do: Map.from_struct(struct)
  def a_one(any), do: any

  def b_two(%_module{} = struct), do: Map.from_struct(struct)
  def b_two(any), do: any

  def c_three(%_module{} = struct), do: Map.from_struct(struct)
  def c_three(any), do: any
```

When serializing the data into MyApp.ExampleView, when the value at key: :a_one is a List, the Serialization of the data does not transform a List of Structs into a list of maps with keys of the appropriate transformed form. 

For example, when we configure JSONAPI to camelize:
```
config :jsonapi, field_transformation: :camelize
```
we expect all the keys in the top-level example map as well as the nested listed data to have keys that are camelized.

Sample input data:
`%MyStruct{a_one: "Data 1", b_two: "Data 2", c_three: "Data 3"}`

MyApp.ExampleView output:
```
%{
  data: %{
    attributes: %{
      "aOne" => "Data 1", "bTwo" => "Data 2", "cThree" => "Data 3" 
    },
    type: "exampleView"
  }
}
```

Sample input data:
```
%MyStruct{
  a_one: [
    %MyOtherStruct{d_one: "Data 4", e_two: "Data 5", f_three: "Data 6"},
    %MyOtherStruct{d_one: "Data 7", e_two: "Data 8", f_three: "Data 9"}
  ], 
  b_two: "Data 2", 
  c_three: "Data 3"
}
```
MyApp.ExampleView actual output:
```
%{
  data: %{
    attributes: %{
      "aOne" => [
        %MyOtherStruct{d_one: "Data 4", e_two: "Data 5", f_three: "Data 6"},
        %MyOtherStruct{d_one: "Data 7", e_two: "Data 8", f_three: "Data 9"}
      ],
      "bTwo" => "Data 2", 
      "cThree" => "Data 3" 
    },
    type: "exampleView"
  }
}
```
MyApp.ExampleView expected output:
```
%{
  "data" => %{
    "attributes" => %{
      "aOne" => [
        %{"dOne" => "Data 4", "eTwo" => "Data 5", "fThree" => "Data 6" },
[...]

PR #219, opened 2020-01-30
Transform keys in lists

Meldanor:
This PR allows to transform keys in maps that are inside a list. 

Solves #217


doomspork:
@jeregrine / @jherdman anything else you'd like to get into the next release along with this or should we just cut it?

jeregrine:
Trust your judgment! Cut a new release whenever.

PR #222, opened 2020-02-25
Deserializer plug

### LINKED ISSUE #210: Deserializer
Hi,

The Serializer works fine. but I would like to know if there is any plan to implement Deserializer module?, If not, What do you recommend to use for it?

snewcomer:
close #210


lib/jsonapi/utils/param_parser.ex
```diff
@@ -0,0 +1,31 @@
...
+defimpl JSONAPI.Utils.ParamParser, for: Any do
+  def parse(data), do: data
+end
+
+defimpl JSONAPI.Utils.ParamParser, for: List do
+  def parse(list), do: Enum.map(list, &JSONAPI.Utils.ParamParser.parse/1)
+end
+
+# Pass built in data types through
+defimpl JSONAPI.Utils.ParamParser,
```
doomspork: Do we need this implementation or the one below? Since we're using `@fallback_to_any` and all of these just pass though, is it necessary?

test/jsonapi/plugs/deserializer_test.exs
```diff
@@ -0,0 +1,105 @@
+defmodule JSONAPI.DeserializerTest do
+  use ExUnit.Case
+  use Plug.Test
+
+  defmodule ExamplePlug do
+    use Plug.Builder
+    plug(Plug.Parsers, parsers: [:json], json_decoder: Jason)
```
doomspork: Style: I don't think we should use parens around plug calls because they exclude those in their own documentation (and formatter config)

snewcomer:
Welp you are right.  My too quick of a take before I started this PR was it was for query params but I didn't take it to it's logical conclusion.  

I'll close this and see if I can help with some docs if need be!

snewcomer:
Reopening to explore the reverse of serializing...

README.md
```diff
@@ -128,6 +128,20 @@ Transforming fields requires two steps:
    end
    ```
 
+3. Deserializing *incoming* body params requires you add the
+   `JSONAPI.Deserializer` Plug to your API's pipeline. Note that the deserializer
+   expects the same casing for your *outgoing* params as your *incoming* params.
```
jherdman: I find this description a tad confusing. The way I understand this plug it sort of "flattens" the parts of a resource object to make it more convenient to use, especially when creating or updating records. The description, as is, has some overlap with the language used by the `UnderscoreParmeters` section, and has overlap with the sample code used by the `UnderscoreParamters` pipeline.

How about this:

> `JSONAPI.Deserializer` is a plug designed to make a JSON:API resource object more convenient to work with when creating or updating resources. This plug works by taking the resource object format and flattening it into an easier to manipulate Map.
>
> Your pipeline in a Phoenix app might look something like this:
> ```elixir
> pipeline :api do
>   plug JSONAPI.EnsureSpec # not required, but recommended for Deserializer
>   plug JSONAPI.Deserializer
> end
> ```

lib/jsonapi/plugs/deserializer.ex
```diff
@@ -0,0 +1,66 @@
+defmodule JSONAPI.Deserializer do
+  @moduledoc """
+  This plug "deserializes" params for ease of use when casting to changesets.
```
jherdman: Assuming my above stated understanding is correct, it may be prudent to use the word "flatten" instead of "deserialize".

lib/jsonapi/utils/data_to_params.ex
```diff
@@ -0,0 +1,80 @@
+defmodule JSONAPI.Utils.DataToParams do
+  @moduledoc ~S"""
+  Converts params in the JSON api format into flat params convenient for
```
jherdman:
```suggestion
  Converts a Map representation of the JSON:API resource object format into a flat Map convenient for
```

test/jsonapi/plugs/deserializer_test.exs
```diff
@@ -0,0 +1,183 @@
...
+    plug Plug.Parsers, parsers: [:json], json_decoder: Jason
+    plug JSONAPI.Deserializer
+    plug :return
+
+    def return(conn, _opts) do
+      send_resp(conn, 200, "success")
+    end
+  end
+
+  @ct "application/vnd.api+json"
```
jherdman: You may just want to reference, or alias, `JSONAPI.mime_type` so that we don't spread this string throughout the code.

test/jsonapi/plugs/deserializer_test.exs
```diff
@@ -0,0 +1,183 @@
...
+        |> put_req_header("content-type", @ct)
+        |> put_req_header("accept", @ct)
+
+      result = ExampleCamelCasePlug.call(conn, [])
+      assert result.params["someNonsense"] == true
+      assert result.params["someMap"]["nested_key"] == true
+      assert result.params["bazId"] == "2"
+    end
+  end
+end
```
jherdman: This is really good, but you may want to add a has-many relationship example too.

README.md
snewcomer: Yes reading again and you are right!  Great suggestion!

lib/jsonapi/plugs/deserializer.ex
```diff
@@ -0,0 +1,66 @@
...
+        }
+      }
+
+  are transformed to:
+
+      %{
+        "id" => "1",
+        "type" => "user"
+        "foo-bar" => true,
+        "baz-id" => "2"
```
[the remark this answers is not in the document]
snewcomer: Great idea!  Added a test.  I'm not sure though there is a way to process other than this...

https://github.com/jeregrine/jsonapi/pull/222/commits/738c2b9e085120f194306261d4a70fc8ee161598#diff-ced812e7235b283a4598ca0ae957ecd0R91

Otherwise, down in this file, I have tests for `includes` with many types.

test/jsonapi/plugs/deserializer_test.exs
snewcomer: I have a bunch of examples in `data_to_params_test.exs`. Do you think that is sufficient?

test/utils/data_to_params_test.exs
```diff
@@ -34,6 +34,64 @@ defmodule JSONAPI.DataToParamsTest do
...
+      }
+    }
+
+    result = JSONAPI.Utils.DataToParams.process(incoming)
+
+    assert result == %{
+      "id" => "1",
+      "type" => "user",
+      "foo-bar" => true,
+      "baz-id" => ["2", "3"]
```
jherdman: Doesn't this mean you have a bug? Imagine a system like this... I have a super class called "Car", and I want to create two records, a "Ford" and a "Subaru". The way I'm reading this you've lost the typing information that the user provided — thus in my scenario I wouldn't know that a user was creating a "Subaru" and a "Ford".

snewcomer: Fixed!  Great point.  

https://github.com/jeregrine/jsonapi/pull/222/commits/2481ee93f0dbb30902401e8bc2493d23264cbc4d

https://github.com/jeregrine/jsonapi/pull/222/commits/e4ffe0643bc06f7cc11df9c4d3b059158ae6667e

PR #223, opened 2020-02-28
Improve documentation around UnderscoreParameters deserializers

snewcomer:
@jeregrine Let me know what you think about these improvements.  They are super minor.  

However, one util that I often use during deserialization (in combination with casing params) is flattening params for changeset casting.  Do you think it is worthwhile to provide a plug for this?

```ex
   %{
      "id" => "1",
      "data" => %{
        "attributes" => %{"foo" => "bar", "baz" => "bat"},
        "type" => "user"
      }
    }
```

converted to

```ex
   %{
      "baz" => "bat",
      "foo" => "bar",
      "id" => "1",
      "type" => "user"
    }
```

ref #222

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ Please see its documentation for details.
 
 JSONAPI has recommended in the past the use of dashes (`-`) in place of underscore (`_`) as a
 word separator for document member keys. However, as of [JSON API Spec (v1.1)](https://jsonapi.org/format/1.1/), it is now recommended that member names
-are camelCased. This library provides various configuration options for maximum flexibility.
+are camelCased. This library provides various configuration options for maximum flexibility including serializing outgoing parameters and deserializing incoming paramenters.
 
 Transforming fields requires two steps:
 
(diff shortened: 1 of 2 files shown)


jherdman:
One thing I really like about having attributes contained within their own map is that it becomes unambiguous what they are. It also means we need not contend or think about nested resources and the like.

snewcomer:
I have a feeling this is what #210 was actually getting at.  The reverse of serializing outgoing params and this is something I have seen a few times (usually hand-rolled).

So a part of incoming params for a POST request would be like so...

```
    %{
      "id" => "1",
      "data" => %{
        "attributes" => %{"foo" => "bar"},
        "relationships" => %{
          "baz" => %{"data" => %{"id" => "2", "type" => "baz"}}
        },
        "type" => "user"
      }
    }
```

and and would be deserialized like so and you can match on a controller action...

```
    %{
      "baz_id" => "2",
      "foo" => "bar",
      "id" => "1",
      "type" => "user"
    }
```

Of course, a plug would still allow those that like "attributes" contained in their own map!

PR #227, opened 2020-03-17
Add caching of build and mix plts

lucacorti:


lucacorti:
@jherdman I did a mistake in the first commit, but apparently subsequent commits on a PR don't trigger a rebuild.

lucacorti:
@jherdman I restored elixir 1.6.6. As you can see from the last workflow run build time is down from ~16m to ~2m, so the cache is doing its job.

lucacorti:
As a bonus tests and dialyzer now run on OTP 21 and 22.

PR #231, opened 2020-06-04
Fix failing test

astery:
Conn converts param values to string, so assert should expect stringified values.


jherdman:
I think the test fail you're seeing is unrelated, and I'll try to dig into that at some point. That said, master is currently green. Can you help me understand how this test is failing for you?

astery:
```
$ mix -v
Erlang/OTP 22 [erts-10.6.4] [source] [64-bit] [smp:12:12] [ds:12:12:10] [async-threads:1] [hipe]

Mix 1.9.4 (compiled with Erlang/OTP 22)
```

plug 1.9.0 in mix.lock

astery:
Sorry, plug version is 1.10.1

PR #234, opened 2020-07-30
Resolves a Weird Fail on Master

jherdman:


jherdman:
This works, but I'm not sure how I feel about yanking 1.6 support. If we're okay with this, we ought to consider this commit part of a breaking change that will (eventually) necessitate a major release.

jeregrine:
I'm okay dropping 1.6, phoenix only goes so far as 1.9 and people can always grab an older release.

Issue #235, opened 2020-09-04
Multiple values for same filter


CostantiniMatteo:
The jsonapi spec states that multiple filter values can be combined in a comma-separated list. For example:

```
GET /comments?filter[post]=1,2 HTTP/1.1
```

However, I noticed that the library returns a string `"1,2"` instead of splitting it and returning `["1", "2"]`.

Is this by design, and therefore such cases needs to be handled manually by splitting the string?
I would be happy to implement it, if the change is welcome.

Thank you!

jherdman:
Definitely a welcome change.

PR #236, opened 2020-09-04
Add support for comma-separated filter values

CostantiniMatteo:
Implements #235

### DIFF
--- a/test/jsonapi/plugs/query_parser_test.exs
+++ b/test/jsonapi/plugs/query_parser_test.exs
@@ -67,6 +67,12 @@ defmodule JSONAPI.QueryParserTest do
     assert filter[:name] == "jason"
   end
 
+  test "parse_filter/2 handles multiple comma-separated filter values" do
+    config = struct(Config, opts: [filter: ~w(name)], view: MyView)
+    filter = parse_filter(config, %{"name" => "jason,api"}).filter
+    assert filter[:name] == ["jason", "api"]
+  end
+
   test "parse_filter/2 raises on invalid filters" do
     config = struct(Config, opts: [], view: MyView)
 
--- a/test/jsonapi/plugs/query_parser_test.exs
+++ b/test/jsonapi/plugs/query_parser_test.exs
@@ -36,11 +36,10 @@ defmodule JSONAPI.QueryParserTest do
   end
 
   setup do
+    previous_env = System.get_env()
     Application.put_env(:jsonapi, :field_transformation, :underscore)
 
-    on_exit(fn ->
-      Application.delete_env(:jsonapi, :field_transformation)
-    end)
+    on_exit(fn -> System.put_env(previous_env) end)
 
     {:ok, []}
   end
--- a/test/jsonapi/plugs/query_parser_test.exs
+++ b/test/jsonapi/plugs/query_parser_test.exs
@@ -36,11 +36,10 @@ defmodule JSONAPI.QueryParserTest do
   end
 
   setup do
+    previous_env = System.get_env()
     Application.put_env(:jsonapi, :field_transformation, :underscore)
 
-    on_exit(fn ->
-      Application.delete_env(:jsonapi, :field_transformation)
-    end)
+    on_exit(fn -> System.put_env(previous_env) end)
 
     {:ok, []}
   end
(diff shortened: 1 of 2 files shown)


jherdman (dismissed):

jherdman:
LGTM. I need to wrap my head around how to release this. I _think_ it's a breaking change despite being what is pretty obviously a bug in the initial implementation. I'd be curious to hear what other maintainers think.

cc @jeregrine @doomspork

CostantiniMatteo:
I've re-read the documentation and I've made a mistake. The part I quoted in the issue is from the [recommendations page](https://jsonapi.org/recommendations/) so it's not supposed to be the default jsonapi behaviour.

Maybe this could be a configurable behaviour?

jherdman:
I'm pretty certain this is the kind of behaviour users of this library would expect. My inclination is to merge this as a bug fix and call it out in the changelog.

CostantiniMatteo:
I do agree with you, I would expect this behaviour, however this is not jsonapi standard.

I was discussing this with @lucacorti, which pointed out that this is only a recommendation's example, and one may want to support a different a strategy for filtering, or may want to use a comma in the filter value.

Isn't it better to at least offer a way to prevent this behaviour via configuration?

jherdman:
You've convinced me: configuration is probably the way to go here.

CostantiniMatteo:
Hey @jherdman, sorry I was off the grid.

I've updated the PR to make the parsing configurable. Let me know what you think.

I don't particularly like the config name, and how I handled the tests (without sacrificing concurrency).
Since they run asynchronously, tests at line 70 and 77 may change the same configuration and one of the two may fail.
The problem may still arise with the current implementation, although much less likely.

test/jsonapi/plugs/query_parser_test.exs
```diff
@@ -67,12 +67,20 @@ defmodule JSONAPI.QueryParserTest do
     assert filter[:name] == "jason"
   end
 
-  test "parse_filter/2 handles multiple comma-separated filter values" do
+  test "parse_filter/2 handles multiple comma-separated filter values if configured" do
+    Application.put_env(:jsonapi, :filter_values_separator, ";")
```
jherdman, line 71: Should this be cleaned up in an [`on_exit`](https://hexdocs.pm/ex_unit/master/ExUnit.Callbacks.html#on_exit/2) hook?

jherdman:
> Hey @jherdman, sorry I was off the grid.

Never sweat about this. F/OSS is done on a volunteer basis for the vast majority of people. Any and all contributions are much appreciated!

test/jsonapi/plugs/query_parser_test.exs
CostantiniMatteo: Absolutely! I've restored a previous commit where the whole env is backed up before tests and restored on exit.
This way we are sure we won't forget to clean up other variables.

CostantiniMatteo:
Could we take the chance to also merge this one among #245 and make a new release? We ended up not using this feature so I forgot about it, but it would still be useful. What do you think?

lucacorti:
@CostantiniMatteo I think merging this would be a mistake as this is really constraining what the spec allows in term of filtering: https://jsonapi.org/format/#fetching-filtering

The spec  specifies that the `filter` parameter is used for filtering and its content are entirely up to implementation specific strategy. Processing the `filter` query param with some predefined logic is really limiting in terms of what can be implemented. Maybe we could explore some delegation to a behaviour like what I implemented earlier for pagination, or leave it entirely to the library user to parse its content?

CostantiniMatteo:
The processing is optional and needs to be configured so it's not limiting and does not break the existing behaviour.

Maybe it can be refactored to accept a module used to parse the `filter` content, and include a parser with support for comma-separated values, instead of the current hard-coded solution.

lucacorti:
You can't implement all possible user needs on this, so extensible is arguably better than any built in logic.

doomspork:
@jeregrine / @jherdman / @lucacorti / @CostantiniMatteo where do we stand with this PR? It looks like it was approved but there's been some discussion about whether this is the change we want.

lucacorti:
@doomspork @CostantiniMatteo as I said I think this would be better addressed with an extensible approach instead of configurable baked in logic.

PR #239, opened 2020-10-22
Allow serialize List value

opodartho:

PR #245, opened 2021-04-22
Fix/string expand fields on leaves

CostantiniMatteo:
Hi guys! I'm opening this PR this fix what I think it's a bug.

When calling `JSONAPI.Utils.String.expand_fields`, the current behaviour is to always apply the transformation function to every element of a list. However, when the list is the value of a map, such strings are actual values and should be left as-is.

For example:
```
> expand_fields(%{"attributes" => %{"corgiName" => ["Wardel"]}}, &underscore/1)
%{"attributes" => %{"corgi_name" => ["wardel"]}}
```

Here `Wardel` is not a field which should be recased, but an actual value that should not change.

If the current implementation is expected behaviour, feel free to close this PR.

Thanks!

### DIFF
--- a/lib/jsonapi/utils/string.ex
+++ b/lib/jsonapi/utils/string.ex
@@ -168,6 +168,9 @@ defmodule JSONAPI.Utils.String do
       iex> expand_fields(%{"attributes" => %{"corgiName" => ["Wardel"]}}, &underscore/1)
       %{"attributes" => %{"corgi_name" => ["Wardel"]}}
 
+      iex> expand_fields(%{"attributes" => %{"someField" => ["SomeValue", %{"nestedField" => "Value"}]}}, &underscore/1)
+      %{"attributes" => %{"some_field" => ["SomeValue", %{"nested_field" => "Value"}]}}
+
       iex> expand_fields([%{"fooBar" => "a"}, %{"fooBar" => "b"}], &underscore/1)
       [%{"foo_bar" => "a"}, %{"foo_bar" => "b"}]
 
--- a/lib/jsonapi/utils/string.ex
+++ b/lib/jsonapi/utils/string.ex
@@ -200,7 +200,6 @@ defmodule JSONAPI.Utils.String do
     {fun.(key), expand_fields(value, fun)}
   end
 
-  @spec expand_fields(tuple, function) :: tuple
   def expand_fields({key, value}, fun) when is_list(value) do
     {fun.(key), maybe_expand_fields(value, fun)}
   end
--- a/lib/jsonapi/utils/string.ex
+++ b/lib/jsonapi/utils/string.ex
@@ -165,6 +165,9 @@ defmodule JSONAPI.Utils.String do
       iex> expand_fields(%{"attributes" => %{"corgiName" => "Wardel"}}, &underscore/1)
       %{"attributes" => %{"corgi_name" => "Wardel"}}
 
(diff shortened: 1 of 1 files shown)


CostantiniMatteo:
I just saw that #239 solves the same problem. The difference between the two is that in this PR objects in lists are transformed, while in the other they left as-is.

.github/workflows/ci.yml
```diff
@@ -21,7 +21,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/setup-elixir@v1
+      - uses: erlef/setup-beam@v1
```
doomspork, line 24: What does this change get us? I'm not sure this is in scope for the propose PR.

doomspork (commented):
Thanks for the PR @CostantiniMatteo. I'm not sure how #245 slipped through my notifications. I'm going to ping the other maintainers so we can get eyes on these 2 and decide which to merge.

.github/workflows/ci.yml
jherdman: The change is for a non-deprecated action to set up Elixir with Github Actions. It's a good change, though I'm not sure it should be part of this PR.

jherdman (commented):
IMHO the simpler PR is better, and we update CI in a different PR.

.github/workflows/ci.yml
CostantiniMatteo: You’re right, I can open a new PR with just this change if you like.

CostantiniMatteo:
> IMHO the simpler PR is better, and we update CI in a different PR.

In general I agree on choosing the simpler one. Just note  however that, while both PR solve the same problem, sometimes results are different (when a value is a list which contains an object). So it might also be worth taking this into account.

The following map:
```
%{someList: [“someValue”, %{someField: “someOtherValue”}]}
```

With this PR results in:
```
%{“some_list” => [“someValue”, %{“some_field” => “someOtherValue”}]}
```

And in this other map with #239:
```
%{“some_list” => [“someValue”, %{“someField” => “someOtherValue”}]}
```

I’m not sure about which choice is better. I didn’t think about leaving object values unchanged but it seems reasonable now.

I’m writing this on mobile, apologies for any mistake.

opodartho:
@CostantiniMatteo As this and #239 solve same issue but changing the keys inside List seems more reasonable.

.github/workflows/ci.yml
doomspork: @CostantiniMatteo can you pull this out into a separate PR and then update this one? I think both should be good to go.

CostantiniMatteo:
I've opened #246 to fix the CI issues and cleaned-up this. Let me know if there's anything else I can do!

lucacorti:
@CostantiniMatteo Can you update your PR with origin master so this is mergeable?

lucacorti:
@jherdman @doomspork Is this ok?

jherdman:
I believe this is OK, but I'm a bit out of the loop with this project as of late. I defer to @doomspork

doomspork (approved):
This looks good to me @snewcomer any thoughts before we merge this?

lucacorti:
@jherdman @doomspork @snewcomer Thanks!

We are using the library at https://github.com/dottori-it and are currently consuming it from git since we need a few improvements made since 1.3.0.

If we can help with maintenance please let us know ❤️

doomspork:
Always wonderful to learn about folks using these libraries 🎉 We're definitely open to help supporting the library 😁

PR #250, opened 2021-06-02
Run transform function only when the value is string or atom

aproxacs:
Found that it raises an error when the value is integer array. 
```
iex> alias JSONAPI.Utils.String, as: JString
iex> JString.expand_fields([12], &JString.camelize/1)
** (FunctionClauseError) no function clause matching in JSONAPI.Utils.String.camelize/1    
...
```

This PR is to fix this error by only running transform function when it is atom or string.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+- package-ecosystem: mix
+  directory: "/"
+  schedule:
+    interval: weekly
+  open-pull-requests-limit: 999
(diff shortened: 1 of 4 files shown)


doomspork:
@aproxacs could take a peek at whether it would be possible to add a test case of this change? It's be great to make sure we do re-introduce this issue down the road.

aproxacs:
I resolved the conflicts and added test.

lucacorti:
@doomspork I think this is googd to go too.

d-staehler:
Could this please be released? Running into a similar issue and this change fixes it :)

doomspork:
@jeregrine I'm going to be pretty busy for the next few days. Could you merge and release this?

lucacorti:
@doomspork can we merge this?

jeregrine (approved):

jeregrine:
> doomspork can we merge this?

Please do not do this. This is a volunteer effort.

PR #258, opened 2021-10-15
Drop CI Elixir 1.9

snewcomer:
Tests are broken.  I'm initially just testing our CI pipeline since it seems to be out of date.  We may or may not want these to be dropped.

ref #257


jherdman (approved):
IMHO :shipit: . No need to burn cycles needlessly.

PR #259, opened 2021-10-17
Handle empty includes

snewcomer:
https://jsonapi.org/format/1.1/#fetching-includes

> An empty value indicates that no related resources should be returned.

### Description
ref #257 

This one is a bit more tricky.  If a user specifies a default `:include`, should we also strip that?  I don't think we should.  Generally a default includes means it is absolutely critical to the main resource.  I could definitely be swayed though.

```
author: {MyApp.UserView, :include}
```

In its current state, all this PR does it make sure the request does not error with the following query parameter.

`/posts?include=`

### DIFF
--- a/lib/jsonapi/plugs/query_parser.ex
+++ b/lib/jsonapi/plugs/query_parser.ex
@@ -201,6 +201,7 @@ defmodule JSONAPI.QueryParser do
     includes =
       str
       |> String.split(",")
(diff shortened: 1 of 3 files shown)


snewcomer:
> pretty serious overhaul with respect to relationship handling

@jherdman Could you explain a bit more as it relates to a potential solution?  Would love to hear your opinion a bit more in depth.

PR #262, opened 2021-12-22
chore: improve ci

yordis:


Serabe:
Any update on this PR? Thank you!

davydog187 (changes requested):
I don't like the idea of outsourcing CI to a third party that @beam-community/team does not control

doomspork:
I talked with @yordis about this in Slack, glad to hear we're on the same page @davydog187 👍 

@yordis do you want to close this PR or update things?

yordis:
@davydog187 do you want me to move the same idea to this organization? Just removing friction

doomspork:
Apologizes for the lack of clarity on next steps, I'd like to close this and hold off on the changes for now.

From my perspective `setup-beam` as a standalone action makes sense because it abstracts away the complexity of installing Elixir and its dependencies, but once that action is run the other steps are largely standard mix commands so I'm not sure we gain from abstracting those.

I do think we can clean these actions up, possibly breaking out credo, test, format, and dialyzer into separate steps and once those actions grow in scope and complexity we can revisit this.

PR #266, opened 2022-03-27
Correctly sets port in generated URLs

jwilger:
The url generation functions in JSONAPI.View did not previously allow the port
to be configured or allow the port to be correctly inferred from the Plug.Conn.
This change addresses both issues.

N.B., the default port on a Plug.Conn is set to 0. Since this can't possibly be
a valid port number, the url generation function will replace this with the
default port number for the scheme. This is likely only relevant in the test
cases, however.


doomspork (approved):
Thank you and welcome @jwilger!

PR #278, opened 2022-12-23
Adding ability to control rendering of relationships

theirishpenguin:
This allows us to workaround upstream bug 277, which is logged at https://github.com/beam-community/jsonapi/issues/277 Additionally, it gives full control over rendering the relationship if desired.

```
defmodule MyAppWeb.AuthorView do
  use JSONAPI.View, type: "authors"

  def fields do
    [:name]
  end

  # The addition of this new hook method allows us to override the rendering of the relationship.
  # In the below contrived example, we return `nil` for the relationship data if the id of the
  # resources is 37. Otherwise it passes through the data unchanged. It is important to be able to
  # return nil because it lets code higher up the call stack - encode_rel_data/2 - to cleanly detect
  # the nil and ultimately surface a { data: null } for the relationship in the response, which works
  # around https://github.com/beam-community/jsonapi/issues/277
  #
  def relationship_data(data) do
    case data.id do
      37 -> nil
    _ -> data
    end
  end
end
```

### DIFF
--- a/lib/jsonapi/serializer.ex
+++ b/lib/jsonapi/serializer.ex
@@ -153,11 +153,16 @@ defmodule JSONAPI.Serializer do
 
   @spec encode_relation(tuple()) :: map()
   def encode_relation({rel_view, rel_data, _rel_url, _conn} = info) do
+    encoded_data = encode_rel_data(rel_view, rel_data)
+
     data = %{
-      data: encode_rel_data(rel_view, rel_data)
+      data: encoded_data
     }
 
-    merge_related_links(data, info, remove_links?())
+    case encoded_data do
+    nil -> data
+    _ -> merge_related_links(data, info, remove_links?())
+    end
   end
 
   defp merge_base_links(%{links: links} = doc, data, view, conn) do
@@ -202,10 +207,17 @@ defmodule JSONAPI.Serializer do
   end
 
   def encode_rel_data(view, data) do
-    %{
-      type: view.type(),
-      id: view.id(data)
-    }
+    transformed_data = view.relationship_data(data)
+
+    case transformed_data do
+    nil ->
+      nil
+    _ ->
+      %{
+        type: view.type(),
+        id: view.id(transformed_data)
+      }
+    end
   end
 
   # Flatten and unique all the included objects
--- a/lib/jsonapi/view.ex
+++ b/lib/jsonapi/view.ex
@@ -117,6 +117,7 @@ defmodule JSONAPI.View do
 
   @type t :: module()
   @type data :: any()
+  @type transformed_data :: any()
   @type field :: atom()
   @type links :: %{atom() => String.t()}
   @type meta :: %{atom() => String.t()}
@@ -130,6 +131,7 @@ defmodule JSONAPI.View do
   @callback hidden(data()) :: [field()]
   @callback links(data(), Conn.t()) :: links()
   @callback meta(data(), Conn.t()) :: meta() | nil
+  @callback relationship_data(data()) :: transformed_data() | nil
   @callback namespace() :: String.t()
   @callback pagination_links(data(), Conn.t(), Paginator.page(), Paginator.options()) ::
               Paginator.links()
@@ -188,6 +190,9 @@ defmodule JSONAPI.View do
 
       @impl View
       def meta(_data, _conn), do: nil
+
+      @impl View
+      def relationship_data(data), do: data
 
       @impl View
       if @namespace do


lib/jsonapi/serializer.ex
```diff
@@ -153,11 +153,16 @@ defmodule JSONAPI.Serializer do
 
   @spec encode_relation(tuple()) :: map()
   def encode_relation({rel_view, rel_data, _rel_url, _conn} = info) do
+    encoded_data = encode_rel_data(rel_view, rel_data)
```
theirishpenguin, line 156: Here, we capture `encoded_data` so that we can check if it's nil on line 162 in order to suppress rendering of links.

Of course, we will also set the `data:` key of the returned Map.

lib/jsonapi/serializer.ex
```diff
@@ -202,10 +207,17 @@ defmodule JSONAPI.Serializer do
   end
 
   def encode_rel_data(view, data) do
-    %{
-      type: view.type(),
-      id: view.id(data)
-    }
+    transformed_data = view.relationship_data(data)
```
theirishpenguin, line 210: Here we will call the new hook method `relationship_data/1`, which we expect the developer to define - or else a default implementation of the method will be used (defined in view.ex - see later).

lib/jsonapi/serializer.ex
```diff
@@ -202,10 +207,17 @@ defmodule JSONAPI.Serializer do
...
+
+    case transformed_data do
+    nil ->
+      nil
```
theirishpenguin, line 214: Key moment: We can now cleanly surface nils back up the call stack.

lib/jsonapi/view.ex
```diff
@@ -189,6 +191,9 @@ defmodule JSONAPI.View do
       @impl View
       def meta(_data, _conn), do: nil
 
+      @impl View
+      def relationship_data(data), do: data
```
theirishpenguin, line 195: Default behaviour for relationship_data/1 is to simply pass through the data.

mattpolzin:
I would also like to see it become possible to serialize null in this way rather than omitting relationship keys.

I have an alternative implementation to suggest, though. I believe it is equally valid to omit to-one relationships or encode them as `null` according to the spec, and I think there are reasonable semantic differences between the two options, so I think it would be nice to easily control which serialization is used on a per-relationship level.

Therefore, I propose we implement this as an option similar to the existing `:include` flag that tells a particular relationship to be included by default. Right now, you can specify a relationship as follows (taken from the README): `author: {MyApp.UserView, :include}`. I think that supporting `author: {MyApp.UserView, [:include, :serialize_empty]}` (naming TBD) would be a nice way to offer very granular control over this feature.

theirishpenguin:
Thanks @mattpolzin. I'd be in broad agreement with your above suggestions. I am not hugely experienced with this library so yourself and others would be better judges of what approach would be more idiomatic for this library.

**One callout:** I have use cases in which it is useful to have some flexibility in what is considered null in my business model layer (eg. in my description on this ticket above, I have the contrived example where the foreign key is 37 and I want that to be surfaced as null). I can understand if allowing for that kind of use case is considered bad taste by those who contribute and maintain this library. But if it can be catered for then that is a win for my requirements.

mattpolzin:
I’ve got a clarifying question. The issue that spawned this PR was about serializing absent relations as null, but your example about id 37 is a separate use-case, right? In the case of id 37, the relation would not have been null but rather something like `{ “data”: { “id”: “37”, “type”: “something” } }` so you are in some cases interested in serializing missing relations as null and in other cases interested in serializing relations that are not missing as null.

theirishpenguin:
@mattpolzin Yes, exactly. This should really be two separate issues in github:

1. Serializing absent relations as null
2. Serializing relations that are not missing as null

mattpolzin:
> Serializing relations that are not missing as null.

While I can imagine situations where I would want to null-out a relationship that the backend knows about but for some reason the client should not, I wonder whether `relationship_data/1` in a `JSONAPI.View` has the right information available to it to really support that use-case in general. In your example (which I realize you did say was contrived), nulling out is based on a hard-coded `id`. In the general case, I imagine this logic would at least be based on a non-constant value (requiring a database lookup perhaps) or even some business logic that required more than knowing only the `id` and type of the relation (probably requiring knowledge other information about the request or the resource that defines the relationship).

Unless you disagree with that, I'd say that `relationship_data/1` is not equipped to handle that use-case more broadly that in the example this PR description lays out. On the other hand, sometime prior to serialization, there must have been some code that actually performed a lookup (perhaps a database query) to retrieve the resource and its relatives and that code necessarily did have all of the aforementioned context (request parameters, database access, knowledge of both the parent resource and the relation) so that feels like a more appropriate place to null-out a relation.

theirishpenguin:
@mattpolzin I would think that you are broadly correct in your last comment. And maybe, as you say, there may be another part of the stack (outside the view layer) _feels like a more appropriate place to null-out a relation._

I suppose I've just seen many cases, down through the years, where there is a small range of well-known IDs or (or single ID) that should not be sent to the client - for "missing" records or similar, where the database design wants to avoid the use of nulls - eg. https://stackoverflow.com/questions/4336687/how-can-i-avoid-nulls-in-my-database-while-also-representing-missing-data

However, I'm only presenting the counter-argument and not arguing for it :smile:

If I was to argue for it, it would probably be something abstract like - it seems reasonable that the view layer has the ability to suppress the surfacing of a relationship. But, I'm not saying this is a knockdown argument.

In terms of syntax, (full disclosure: I have more of Ruby background than an Elixir background so forgive me if the following example doesn't compile - as I'm more focused on eating Christmas leftovers at the moment than coding on my laptop :rofl:  )...

```
author: {MyApp.UserView, [:include, :serialize_empty, empty_if: fn data -> data.author_id != 1 end]}
```

So, in summary, I think that it (1) Serializing absent relations as null should be implemented and I agree that (2) serializing relations that are not missing as null could be left unimplemented for now, or forever, unless there requests for such a feature or better arguments in favour of it.

Thanks again for the quick replies.

mattpolzin:
> there may be another part of the stack (outside the view layer) that feels like a more appropriate place to null-out a relation.

This is of course one of those situations where I wouldn't dare pretend to have the right answer for another person's codebase, but I'll just mention how I would approach it briefly because the following would be compatible with the current capabilities of this JSON:API library (or any other serialization framework) because it only relies on `Ecto`.

I'd use a filtered relationship on my schema: https://hexdocs.pm/ecto/Ecto.Schema.html#has_many/3-filtering-associations. This filter could hard code some ids or it could use a field in the database (perhaps the relation's table has a `expose_to_client` field)  or maybe you're looking for soft-deletion (perhaps the relation's table has a `is_visible` or `deleted_at` field).

In the example that would mirror the original PR description above,
```elixir
defmodule MyApp.Schema.Author do
  ...
  schema "authors" do
    ...
    has_one :publisher, Publisher,
      where: [id: {:not, 37}]
  end
end
```

theirishpenguin:
Hi @mattpolzin thanks very much for your suggestion. I had look at this option but a couple of things jumped out at me...

* `where: [id: {:not, 37}]` didn't actually work for me (the docs say `where: [id: {:not, nil}]` and, after if failed when I tried an integer ID, I assumed that maybe that ecto functionality only worked with `nil` and not an ID integer - I'm not sure if it was just an issue on my machine)
* I was able to get the equivalent of that syntax working with `{:fragment, expr}` which is also listed in those docs, but...
* I noticed that the ecto relationship would be null, but that the foreign key ID (ie. author_id id 37 on my Book struct) would still be 37. So if that struct is passed around then there is a contraction between the author_id (which is not null) and the author relationship (which is null). Ideally, I don't think I want that contradiction - so, given the struct will contain an author_id of 37, I want it to keep pointing at my non-null "placeholder" record
* That's why I only want the relationship to go to null as late as possible (ie. when surfacing the relationship in the API - as there will be no author_id of 37 to speak of at that moment in time)
* I could add a layer to null out both the author_id and the relationship at the same time - a layer that would exist between the ecto layer and the jsonapi view layer - but that felt like it might be over-engineering things
* That's why the best tradeoff for me seemed to be to have the existing jsonapi view layer (which avoids me having to create another layer) offer me some control over how to surface the relationship (or not)

Again, not arguing for the case for this feature - just explaining my thinking. I could be thinking about this all wrong - either way, this conversation is useful learnings for me, thanks!

mattpolzin:
Thanks for going back and forth with me on this.

The distinction between a nulled out relationship vs a non-null Id is interesting. I guess I’ve only ever considered an relational field to be an attribute (`author_id`) or a relationship (`author`) but never access the same field as both in my code (I surface it as an attribute _or_ a relationship in my Ecto Schema) so I’ve never had to worry about a association filter causing the relationship to be nil while the id property is not.

I’m only one person so I’ll leave the door open for others to voice their opinions but specifically the feature to null out relationships inside views using only the data property of the relationship itself feels like a narrow enough use-case (e.g. hard coding certain ids to be null) that id like to see others have need for something similar before adopting it into the library. Not just because if more people need it then it’s worth having, but also because others might have subtly different but very similar needs that would influence how the feature was designed. 

I think it’s worth the effort to add an ability to control whether null relations are serialized as null or omitted, though. I’ve seen the use cases for that distinction numerous times before.

theirishpenguin:
Hi @mattpolzin, I agree with your points above. If I get a chance to rework this to follow the approach you noted at https://github.com/beam-community/jsonapi/pull/278#issuecomment-1364107901 then I will note it here. (I may not get around to that for quite a while depending on my schedule).

In the meantime, happy new year and thanks for all your discussion above!

theirishpenguin:
I'm not even close to getting time to implement the approach noted above. And I can currently work around the issue. Given a fresh PR will clean content in it would probably be better anyway, I plan close this PR soon (unless there are objections).

mattpolzin:
Sounds good. Thanks!

theirishpenguin:
Thanks :+1: Closing PR as promised.

Issue #274, opened 2022-10-26
Publish New Version on Hex


TylerPachal:
I was surprised to see that the latest version of this library to be published [on Hex is `1.3.0` from Mar 21, 2020](https://hex.pm/packages/jsonapi).  There have been quite a few commits since then, and according to the `mix.exs` file we are now on version `1.4.0`.  Could `1.4.0` be published to Hex?

Thanks!

doomspork:
We will get a new version released here shortly @TylerPachal! I'm going to work with @btkostner to setup workflows to ensure we keep the latest and greatest on Hex.

yordis:
More than happy to help with the CI situation if you want me to

TylerPachal:
@doomspork I am closing this now, since version 1.4.0 is now available on Hex 🥳

PR #280, opened 2022-12-28
Push new releases to hex

doomspork:
This aims to address #274. cc @yordis @btkostner 

Long term I think it's worth discussing the usage of something like Conventional Commits + Release Please like @btkostner has configured for us at @stordco, but this should allow us to immediately address the new to cut releases to hex and do some based on GitHub releases rather than updates to the `VERSION` file as I did with other libs under beam-community.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,23 @@
+---
+name: Publish to Hex.pm
+
+on:
+  release:
+    types:
+      - released
+  workflow_dispatch: {}
+
+jobs:
+  Publish:
+    runs-on: ubuntu-20.04
+    env:
+      HEX_API_KEY: ${{ secrets.HEXPM_SECRET }}
+    steps:
+      - uses: actions/checkout@v2
+      - uses: erlef/setup-beam@v1
+        with:
+          otp-version: 25
+          elixir-version: 1.13
+      - run: mix deps.get
+      - run: mix compile --docs
+      - run: mix hex.publish --yes
--- a/README.md
+++ b/README.md
@@ -1,5 +1,11 @@
 # JSONAPI Elixir
 
+[![Build](https://github.com/beam-community/jsonapi/actions/workflows/ci.yml/badge.svg)](https://github.com/beam-community/jsonapi/actions/workflows/ci.yml)
+[![Hex.pm version](https://img.shields.io/hexpm/v/jsonapi.svg)](https://hex.pm/packages/jsonapi)
+[![Hex.pm downloads](https://img.shields.io/hexpm/dt/jsonapi.svg)](https://hex.pm/packages/jsonapi)
+[![Hex.pm weekly downloads](https://img.shields.io/hexpm/dw/jsonapi.svg)](https://hex.pm/packages/jsonapi)
+[![Hex.pm daily downloads](https://img.shields.io/hexpm/dd/jsonapi.svg)](https://hex.pm/packages/jsonapi)
+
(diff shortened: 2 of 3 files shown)


mix.exs
```diff
@@ -64,10 +64,15 @@ defmodule JSONAPI.Mixfile do
         "Mitchell Henke",
         "Jake Robers",
         "Sean Callan",
-        "James Herdman"
+        "James Herdman",
+        "Mathew Polzin"
```
doomspork, line 68: Added @mattpolzin as a maintainer since he will be the primary driver for this lib 🎉

mix.exs
```diff
@@ -64,10 +64,15 @@ defmodule JSONAPI.Mixfile do
...
       ],
       licenses: ["MIT"],
-      links: %{github: "https://github.com/jeregrine/jsonapi", docs: "http://hexdocs.pm/jsonapi/"}
+      files: ["lib", "mix.exs", "README*", "LICENSE*"],
+      links: %{
+        github: "https://github.com/beam-community/jsonapi",
```
doomspork, line 73: Updated the GitHub URL now that this has been transferred into the beam-community org.

mix.exs
```diff
@@ -64,10 +64,15 @@ defmodule JSONAPI.Mixfile do
         "Mitchell Henke",
         "Jake Robers",
         "Sean Callan",
-        "James Herdman"
+        "James Herdman",
+        "Mathew Polzin"
       ],
       licenses: ["MIT"],
-      links: %{github: "https://github.com/jeregrine/jsonapi", docs: "http://hexdocs.pm/jsonapi/"}
+      files: ["lib", "mix.exs", "README*", "LICENSE*"],
```
doomspork, line 71: Am I missing anything here? 🤔

README.md
```diff
@@ -1,5 +1,11 @@
 # JSONAPI Elixir
 
+[![Build](https://github.com/beam-community/jsonapi/actions/workflows/ci.yml/badge.svg)](https://github.com/beam-community/jsonapi/actions/workflows/ci.yml)
+[![Hex.pm version](https://img.shields.io/hexpm/v/jsonapi.svg)](https://hex.pm/packages/jsonapi)
+[![Hex.pm downloads](https://img.shields.io/hexpm/dt/jsonapi.svg)](https://hex.pm/packages/jsonapi)
+[![Hex.pm weekly downloads](https://img.shields.io/hexpm/dw/jsonapi.svg)](https://hex.pm/packages/jsonapi)
+[![Hex.pm daily downloads](https://img.shields.io/hexpm/dd/jsonapi.svg)](https://hex.pm/packages/jsonapi)
```
doomspork, line 7: Added build status and Hex.pm badges to README. Other changes are from prettier.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,23 @@
+---
+name: Publish to Hex.pm
+
+on:
+  release:
+    types:
+      - released
```
doomspork, line 7: @btkostner do we _need_ to support pre-releases here?

mix.exs
mattpolzin: I'm pretty new to publishing to Hex, but do we _want_ to override the default given the description "Defaults to standard project directories, so you usually don't need to set this property."

doomspork: Good callout @mattpolzin! I went to see what the latest and great defaults are and found these: https://github.com/hexpm/hex/blob/d78a23f617dda568cc11642ada459cc3b2683787/lib/hex/package.ex#L5-L6

Looks like we're covered 👍

mattpolzin (approved):
Seems like a big win for efficiency to me!

PR #281, opened 2022-12-28
Update CI workflow to use main

doomspork:
Moving all the beam-community repos to `main` instead of `master` to align with larger industry shift. Updated CI and prettier applied.


mattpolzin (approved):
Awesome. This was on my mind as well.

btkostner (approved):

PR #282, opened 2023-01-04
Integration between UnderscoreParameters and QueryParser

TylerPachal:
The Problem:

If you are using the `UnderscoreParameters` plug, followed by filtering with the `QueryParser` plug, the filtering will not work.  Here is an example:

```elixir
# router.ex

pipeline :api do
  plug JSONAPI.EnsureSpec
  plug JSONAPI.Deserializer
  plug JSONAPI.UnderscoreParameters
end
```

```elixir
# foobar_controller.ex

plug JSONAPI.QueryParser,
  filter: ~w(underscored_field_name),
  view: FoobarView
```

The expectation is that I can make the following request, and have the dasherized field in my query string be converted to underscores by the time it hits the controller:

```
GET /foobar?filter[underscored-field-name]=whatever
```

**The reason it doesn't work is** because the `UnderscoreParameters` plug replaces the `params` field of the `Plug.Conn`, but the subsequent `QueryParser` is checking in the `query_params` field instead.

---

The solution:

This PR adds support for an opt-in option on the `UnderscoreParameters` plug called `replace_query_params`.  It is opt-in so that nothing breaks for existing users.  When you set the option to `true` the `query_params` will be replaced on the `Plug.Conn`, not just the `params`.  

```elixir
# router.ex

pipeline :api do
  plug JSONAPI.EnsureSpec
  plug JSONAPI.Deserializer
  plug JSONAPI.UnderscoreParameters, replace_query_params: true # <--- this will make things work
end
```

### DIFF
--- a/lib/jsonapi/plugs/query_parser.ex
+++ b/lib/jsonapi/plugs/query_parser.ex
@@ -7,9 +7,9 @@ defmodule JSONAPI.QueryParser do
   import JSONAPI.Utils.String, only: [underscore: 1]
 
   @moduledoc """
-  Implements a fully JSONAPI V1 spec for parsing a complex query string and
-  returning Elixir datastructures. The purpose is to validate and encode incoming
-  queries and fail quickly.
+  Implements a fully JSONAPI V1 spec for parsing a complex query string via the
+  `query_params` field from a `Plug.Conn` struct and returning Elixir datastructures.
+  The purpose is to validate and encode incoming queries and fail quickly.
 
   Primarialy this handles:
     * [sorts](http://jsonapi.org/format/#fetching-sorting)
@@ -72,7 +72,8 @@ defmodule JSONAPI.QueryParser do
 
   Note that if your API is returning dasherized fields (e.g. `"dog-breed": "Corgi"`)
   we recommend that you include the `JSONAPI.UnderscoreParameters` Plug in your
-  API's pipeline. This will underscore fields for easier operations in your code.
+  API's pipeline with the `replace_query_params` option set to `true`. This will
+  underscore fields for easier operations in your code.
 
   For more details please see `JSONAPI.UnderscoreParameters`.
   """
--- a/lib/jsonapi/plugs/underscore_parameters.ex
+++ b/lib/jsonapi/plugs/underscore_parameters.ex
@@ -88,16 +88,16 @@ defmodule JSONAPI.UnderscoreParameters do
   def call(%Plug.Conn{params: params} = conn, opts) do
     content_type = get_req_header(conn, "content-type")
 
-    conn =
-      if opts[:replace_query_params] == true do
-        query_params = fetch_query_params(conn).query_params
-        new_query_params = JString.expand_fields(query_params, &JString.underscore/1)
-        Map.put(conn, :query_params, new_query_params)
-      else
-        conn
-      end
-
     if JSONAPI.mime_type() in content_type do
+      conn =
+        if opts[:replace_query_params] do
+          query_params = fetch_query_params(conn).query_params
+          new_query_params = JString.expand_fields(query_params, &JString.underscore/1)
+          Map.put(conn, :query_params, new_query_params)
+        else
+          conn
+        end
+
       new_params = JString.expand_fields(params, &JString.underscore/1)
       Map.put(conn, :params, new_params)
     else
--- a/lib/jsonapi/plugs/underscore_parameters.ex
+++ b/lib/jsonapi/plugs/underscore_parameters.ex
@@ -88,16 +88,16 @@ defmodule JSONAPI.UnderscoreParameters do
   def call(%Plug.Conn{params: params} = conn, opts) do
     content_type = get_req_header(conn, "content-type")
 
-    if JSONAPI.mime_type() in content_type do
-      conn =
-        if opts[:replace_query_params] do
-          query_params = fetch_query_params(conn).query_params
-          new_query_params = JString.expand_fields(query_params, &JString.underscore/1)
-          Map.put(conn, :query_params, new_query_params)
-        else
-          conn
-        end
+    conn =
+      if opts[:replace_query_params] == true do
+        query_params = fetch_query_params(conn).query_params
+        new_query_params = JString.expand_fields(query_params, &JString.underscore/1)
+        Map.put(conn, :query_params, new_query_params)
+      else
+        conn
+      end
 
+    if JSONAPI.mime_type() in content_type do
       new_params = JString.expand_fields(params, &JString.underscore/1)
       Map.put(conn, :params, new_params)
     else
--- a/lib/jsonapi/plugs/underscore_parameters.ex
+++ b/lib/jsonapi/plugs/underscore_parameters.ex
@@ -1,12 +1,24 @@
 defmodule JSONAPI.UnderscoreParameters do
   @moduledoc """
-  Takes dasherized JSON:API params and deserializes them to underscored params. Add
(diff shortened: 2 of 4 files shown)


lib/jsonapi/plugs/query_parser.ex
```diff
@@ -7,9 +7,9 @@ defmodule JSONAPI.QueryParser do
   import JSONAPI.Utils.String, only: [underscore: 1]
 
   @moduledoc """
-  Implements a fully JSONAPI V1 spec for parsing a complex query string and
-  returning Elixir datastructures. The purpose is to validate and encode incoming
-  queries and fail quickly.
+  Implements a fully JSONAPI V1 spec for parsing a complex query string via the
+  `query_params` field from a `Plug.Conn` struct and returning Elixir datastructures.
```
TylerPachal, line 11: Annotation: I added some clarity to the documentation to emphasize that the this Plug looks at the `query_params` and not the `params`.

test/jsonapi/plugs/query_parser_test.exs
```diff
@@ -151,4 +153,23 @@ defmodule JSONAPI.QueryParserTest do
...
+    config = struct(Config, view: MyView, opts: [filter: ["favorite_food"]])
+
+    conn =
+      conn
+      |> JSONAPI.UnderscoreParameters.call(replace_query_params: true)
+      |> JSONAPI.QueryParser.call(config)
+
+    # Ensure the underscored file name is present in the parsed filters
+    assert [favorite_food: _] = conn.assigns.jsonapi_query.filter
+  end
```
TylerPachal, line 174: Annotation: This is the "main" test, which makes sure that the `UnderscoreParameters` and `QueryParser` plugs will work together.

test/jsonapi/plugs/underscore_parameters_test.exs
```diff
@@ -4,89 +4,123 @@ defmodule JSONAPI.UnderscoreParametersTest do
...
-  test "underscores dasherized data parameters" do
-    params = %{
-      "data" => %{
-        "attributes" => %{
-          "first-name" => "John",
-          "last-name" => "Cleese",
-          "stats" => %{
-            "age" => 45,
-            "dog-name" => "Pedro"
+  describe "call/2" do
```
TylerPachal, line 7: Annotation: I added this `describe` block to separate the `call` tests from the `init` test.  I recommend viewing the diff with Github's `Hide Whitespace` option to make the other changes standout.

lib/jsonapi/plugs/underscore_parameters.ex
```diff
@@ -55,15 +67,38 @@ defmodule JSONAPI.UnderscoreParameters do
   alias JSONAPI.Utils.String, as: JString
 
   @doc false
-  def init(opts), do: opts
+  def init(opts) do
```
mattpolzin, line 70: The following is certainly functional and raising an error for the wrong type is really nice; I'd like to suggest a stylistic change if you don't mind: Using two function clauses, one that matches when `replace_query_params` is not a boolean (that clause can simply raise your error) and then a default clause with the same definition as `init/1` had before to just pass `opts` on in all other cases.

I can say more if you're not sure what I am picturing.

mattpolzin (changes requested):
Nice! One stylistic request.

lib/jsonapi/plugs/underscore_parameters.ex
TylerPachal: Roger that.  I will make the change now.

TylerPachal: Alternatively we could remove the checking in the `init` block.  I don't think it is 100% necessary, but it is a nice-to-have.

mattpolzin: Hmmm, well, I was conflating Elixir's behavior when pattern matching dictionaries with its behavior when pattern matching lists so the solution I pictured is not valid after all.

> because if the opts list ever (accidentally) contained more than one element, the pattern wouldn't match and our check would be skipped...

We could also check the length of the list to throw an error about only supporting one option, but that's not a very sustainable solution either because the second we do add a second option we need to re-evaluate our pattern matching strategy anyway.

Thanks for giving it a try. I think your original code here feels best given the error you are raising is a very nice developer experience.

lib/jsonapi/plugs/underscore_parameters.ex
```diff
@@ -1,13 +1,25 @@
...
   @moduledoc """
-  Takes dasherized JSON:API params and deserializes them to underscored params. Add
+  Takes dasherized JSON:API params and converts them to underscored params. Add
   this to your API's pipeline to aid in dealing with incoming parameters such as query
   params or data.
 
+  By default the newly underscored params will only replace the existing `params` field
+  of the `Plug.Conn` struct, but leave the `query_params` and `body_params` untouched.
+  If you are using the `JSONAPI.QueryParser` and need to also have the `query_params` on
+  the `Plug.Conn` updated, set the `replace_query_params` option to `true`.
```
snewcomer, line 10: Hi!  So I'm trying to piece together the history here.

The original incarnation of the UnderscoreParameters was [this](https://github.com/beam-community/jsonapi/issues/145) - to stop dasherizing fields and work with the QueryParser.   cc @jherdman. 

The original documentation for that PR was....

```
  Given a request like:

       GET /example?filters[dog-breed]=Corgi
```
  
I suspect this was just an implementation oversight.  For incoming params, the benefit of the UnderscoreParameters plug was for convenience. I would assume replacing query-params also brings that convenience.  I get how this might be a breaking change for some rather than a bug fix though.  

We should fully expect it to be the default rather than a config option in the future.

jherdman: Definitely an oversight on my part.

mattpolzin:
I’m not sure I follow your latest commit’s purpose [here](https://github.com/beam-community/jsonapi/pull/282/commits/dfe1dde0a572ecc72fcc04eaf107a642826e571f). 

If I read it correctly, the behavior is:
1. If `replace_query_params` is false, only underscorize in the params map if the content type is JSON:API. 
2. If `replace_query_params` is true, always underscorize in the query_params map in addition to only underscorizing in the params map if the content type is JSON:API. 

This feels like an awkward difference in the treatment of the query_params and params outputs. 

I believe (though this is speculation based on my own intuition) that the idea behind only applying the underscore logic when the content type is JSON:API is to make it so that you can use this plug provided specifically by a JSON:API library in a pipeline that also handles non-JSON:API requests and you won’t have behavior intended only for JSON:API bleed over into other request types.

This last change also seems to bring into scope the conversation over in https://github.com/beam-community/jsonapi/issues/247 where the intention is to remove the need for content type to match JSON:API specifically for GET requests but still check the ACCEPT header in those cases to avoid applying logic when a request is not JSON:API at all.

TylerPachal:
@mattpolzin You interpreted the commit correctly.  I didn't know about https://github.com/beam-community/jsonapi/issues/247, and while my personal opinion is that it is strange to require the `Content-Type` header for `GET` requests (which have no content), I will undo that commit from this PR and raise that issue separately 👍

mattpolzin:
@TylerPachal the prevailing opinion in https://github.com/beam-community/jsonapi/issues/247 aligns with yours: GET requests will not need to have a content type set at all, only an ACCEPT header specifying JSON:API

mattpolzin (approved):
Looking good to me! I'll create a GitHub issue after this merges to make this the default behavior (rendering the new option a no-op) with the next major release of the library per the recommendation Scott left in a comment.

PR #287, opened 2023-01-25
Only underscore field[...] values

TylerPachal:
**The Problem**

If you pass in query parameters like this: `?fields[my-type]=field-abc` and used the `replace_query_params: true` option, then the results would look like:

```elixir
%{"fields" => %{"my_type" => "field-abc"}}
```

However, the results we actually wanted were this:

```elixir
%{"fields" => %{"my-type" => "field_abc"}}
```

Where the type remains unchanged to match the kebab-case type in the view, and the field gets underscored to match the field name in the Elixir codebase:

```elixir
defmodule MyTypeView do
  def type, do: "my-type"
  
  def fields, do: [:field_abc]
end
```

**The Solution**

The fix is to be a little more precise in the `UnderscoreParameters` plug.  If the `fields` query parameter exists, ensure that we do not change the keys, but do change the values.

### DIFF
--- a/lib/jsonapi/plugs/underscore_parameters.ex
+++ b/lib/jsonapi/plugs/underscore_parameters.ex
@@ -89,6 +89,8 @@ defmodule JSONAPI.UnderscoreParameters do
     content_type = get_req_header(conn, "content-type")
 
     if JSONAPI.mime_type() in content_type do
+      # In version 2.0, when this block is no longer conditional and applies every time, ensure
+      # that we apply the same treatment to the query_params and "regular" params.
       conn =
         if opts[:replace_query_params] do
           query_params = fetch_query_params(conn).query_params
--- a/lib/jsonapi/plugs/underscore_parameters.ex
+++ b/lib/jsonapi/plugs/underscore_parameters.ex
@@ -92,7 +92,7 @@ defmodule JSONAPI.UnderscoreParameters do
       conn =
         if opts[:replace_query_params] do
           query_params = fetch_query_params(conn).query_params
-          new_query_params = JString.expand_fields(query_params, &JString.underscore/1)
+          new_query_params = replace_query_params(query_params)
           Map.put(conn, :query_params, new_query_params)
         else
           conn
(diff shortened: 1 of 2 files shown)


lib/jsonapi/plugs/underscore_parameters.ex
```diff
@@ -104,4 +104,19 @@ defmodule JSONAPI.UnderscoreParameters do
...
+
+  defp replace_query_params(query_params) do
+    # Underscore the keys of all of the query parameters
+    underscored_query_params = JString.expand_fields(query_params, &JString.underscore/1)
+
+    # If the fields[...] query parameter is present, only underscore its values, but not its keys
+    case Map.fetch(query_params, "fields") do
+      {:ok, fields} when is_map(fields) ->
+        fields = Map.new(fields, fn {k, v} -> {k, JString.underscore(v)} end)
+        Map.put(underscored_query_params, "fields", fields)
```
TylerPachal, line 116: Annotation: This is the main change.  If `fields` are present in the original `query_params`, then underscore all of the values, and put it with the rest of the `underscored_query_params`.

lib/jsonapi/plugs/underscore_parameters.ex
```diff
@@ -92,7 +92,7 @@ defmodule JSONAPI.UnderscoreParameters do
       conn =
         if opts[:replace_query_params] do
           query_params = fetch_query_params(conn).query_params
-          new_query_params = JString.expand_fields(query_params, &JString.underscore/1)
+          new_query_params = replace_query_params(query_params)
```
mattpolzin, line 95: Should this logic not apply to the `new_params` below?

TylerPachal: Sorry - I don't understand the question.  I created this `replace_query_params` function because the logic was becoming too complex to in-line here.

mattpolzin: My question is whether you intended for the behavior of this plugin for `query_params` to diverge from the behavior of `params` when you modified this line but not line 101 below.

TylerPachal: Gotcha.  Yes that was intentional.  I figured it was safest to put this change behind the `:replace_query_params` option like we did in: https://github.com/beam-community/jsonapi/pull/282.

mattpolzin: Oh, good call. Can you add a code comment to the effect of “when we make this unconditional in the next major version we should apply the same treatment to query_params and params.”

TylerPachal: @mattpolzin Added the comment - let me know if that is sufficient (or feel free to change it!)

mattpolzin (approved):

PR #288, opened 2023-01-30
Include NotLoaded relationships

TylerPachal:
**The Problem**

Currently, it is not possible to view _any_ information about a relationship unless you `include` it:

```
GET  /foobar/25
{
  "data": {
    "attributes": {...},
    "id": "25",
    "links": {
      "self": "https://example.com/foobars/25"
    },
    "relationships": {}, // <----- This is always empty
    "type": "foobars"
  }
}
```

And when you do `include`, it can be expensive as it triggers extra queries.

---

**This PR**

We can attach information we have about a relationships, like the `id`, without needing to include anything or perform any additional DB queries:

```
GET  /foobar/25

{
  "data": {
    "attributes": {...},
    "id": "25",
    "links": {
      "self": "https://example.com/foobars/25"
    },
    "relationships": { // <---- There is now data here about the relationship
      "foo": {
        "id": "354",
        "type": "foos"
      }
    },
    "type": "foobars"
  }
}
```

This is possible because even though the `Foobar` model would have a `NotLoaded` value for `foo`, there should still be a foreign key present on the model that we can use to populate the ID and links:

```elixir
%Foobar{
  foo_id: 354, # <--- We can use this, even though the foo is not loaded
  foo: #Ecto.Association.NotLoaded<association :foo is not loaded>
}
```

---

Discussion Points:

- [ ] Should this be behind a config flag?
- [ ] Should we be more precise when looding foreign key (see code annotation)

### DIFF
--- a/lib/jsonapi/plugs/underscore_parameters.ex
+++ b/lib/jsonapi/plugs/underscore_parameters.ex
@@ -89,6 +89,8 @@ defmodule JSONAPI.UnderscoreParameters do
     content_type = get_req_header(conn, "content-type")
 
     if JSONAPI.mime_type() in content_type do
+      # In version 2.0, when this block is no longer conditional and applies every time, ensure
+      # that we apply the same treatment to the query_params and "regular" params.
       conn =
         if opts[:replace_query_params] do
           query_params = fetch_query_params(conn).query_params


lib/jsonapi/serializer.ex
```diff
@@ -186,26 +197,48 @@ defmodule JSONAPI.Serializer do
...
+      else
+        Map.get(entity, :"#{key}_id")
+      end
+
+    if id do
+      %{
+        type: view.type(),
+        id: id
+      }
+    end
```
TylerPachal, line 241: Annotation: This is the main change, all of the other changes are really just to pass the `entity` and `key` down to this point.

Discussion Point: Line 233 could be improved by using ecto to get the exact name of the foreign key:
```elixir
# Old
Map.get(entity, :"#{key}_id")

# New
%{owner_key: owner_key} = entity.__struct__.__schema__(:association, key)
```

The reason I did not do this is because it broke all of the Serializer tests.  Those tests are using a plain map as their input test data, but in reality I think this would always be a struct backed by an Ecto Schema.  If we want to make that change, we will need to include Ecto as a test dependency, and that was starting to make this PR too large.

lib/jsonapi/serializer.ex
```diff
@@ -79,12 +79,12 @@ defmodule JSONAPI.Serializer do
   @spec encode_relationships(Conn.t(), document(), tuple(), list()) :: tuple()
   def encode_relationships(conn, doc, {view, data, _, _} = view_info, options) do
     view.relationships()
-    |> Enum.filter(&assoc_loaded?(Map.get(data, elem(&1, 0))))
```
TylerPachal, line 82: Annotation: If we want to put this feature behind a config flag, we should place the config flag here, to continue filtering-out unloaded relationships.

PR #289, opened 2023-01-30
Show NotLoaded Relationship IDs

TylerPachal:
**The Problem**

Currently, it is not possible to view _any_ information about a relationship unless you `include` it.  If you do not use the `include` query parameter, the `relationships` will be an empty map, not even showing what relationships are possible:

```
GET  /foobar/25
{
  "data": {
    "attributes": {...},
    "id": "25",
    "links": {
      "self": "https://example.com/foobars/25"
    },
    "relationships": {}, // <----- This is always empty
    "type": "foobars"
  }
}
```

The work around for this is to `include` relationships to check for their existence.  Unfortunately this is expensive because it generates additional `JOIN` queries.

---

**This PR**

Without doing any additional queries, we can attach information we already have about a relationships, like the `id`:

```
GET  /foobar/25

{
  "data": {
    "attributes": {...},
    "id": "25",
    "links": {
      "self": "https://example.com/foobars/25"
    },
    "relationships": { // <---- There is now data here about the relationship
      "foo": {
        "id": "354",
        "type": "foos"
      }
    },
    "type": "foobars"
  }
}
```

This is possible because even though the `Foobar` model would have a `NotLoaded` value for `foo`, there should still be a foreign key present on the model that we can use to populate the ID and links:

```elixir
%Foobar{
  foo_id: 354, # <--- We can use this, even though the foo is not loaded
  foo: #Ecto.Association.NotLoaded<association :foo is not loaded>
}
```

---

Discussion Points:

- [ ] Should this be behind a config flag?
- [ ] Should we be more precise when looding foreign key (see code annotation)

### DIFF
--- a/lib/jsonapi/serializer.ex
+++ b/lib/jsonapi/serializer.ex
@@ -79,15 +79,13 @@ defmodule JSONAPI.Serializer do
   @spec encode_relationships(Conn.t(), document(), tuple(), list()) :: tuple()
   def encode_relationships(conn, doc, {view, data, _, _} = view_info, options) do
     view.relationships()
-    |> Enum.filter(&data_loaded?(Map.get(data, get_data_key(&1))))
-    |> Enum.map_reduce(doc, &build_relationships(conn, view_info, &1, &2, options))
+    |> Enum.map_reduce(doc, &build_relationships(data, conn, view_info, &1, &2, options))
   end
 
-  defp get_data_key(rel_config), do: elem(extrapolate_relationship_config(rel_config), 1)
-
-  @spec build_relationships(Conn.t(), tuple(), term(), term(), module(), tuple(), list()) ::
+  @spec build_relationships(map(), Conn.t(), tuple(), term(), term(), module(), tuple(), list()) ::
           tuple()
   def build_relationships(
+        entity,
         conn,
         {parent_view, parent_data, query_includes, valid_includes},
         relationship_name,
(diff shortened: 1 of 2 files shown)


lib/jsonapi/serializer.ex
```diff
@@ -79,15 +79,13 @@ defmodule JSONAPI.Serializer do
   @spec encode_relationships(Conn.t(), document(), tuple(), list()) :: tuple()
   def encode_relationships(conn, doc, {view, data, _, _} = view_info, options) do
     view.relationships()
-    |> Enum.filter(&data_loaded?(Map.get(data, get_data_key(&1))))
```
TylerPachal, line 82: Discussion Point: Do we want to put the new functionality behind some sort of flag?

```elixir
if options[:auto_include_relationship_ids]
    view.relationships()
    |> Enum.map_reduce(doc, &build_relationships(data, conn, view_info, &1, &2, options))
else
    view.relationships()
    |> Enum.filter(&data_loaded?(Map.get(data, get_data_key(&1))))
    |> Enum.map_reduce(doc, &build_relationships(data, conn, view_info, &1, &2, options))
end

```

lib/jsonapi/serializer.ex
```diff
@@ -233,26 +244,48 @@ defmodule JSONAPI.Serializer do
...
-      type: view.type(),
-      id: view.id(data)
-    }
+  def encode_rel_data(entity, key, view, data) do
+    id =
+      if assoc_loaded?(data) do
+        view.id(data)
+      else
+        Map.get(entity, :"#{key}_id")
+      end
```
TylerPachal, line 281: Annotation: This is the main change.  Instead of just loading the ID if the `data` is loaded, get the ID from the foreign key if the data is not loaded.

Discussion Point: Line 280 could be improved by using Ecto to get the exact name of the foreign key:
```elixir
# Old
Map.get(entity, :"#{key}_id")

# New
%{owner_key: owner_key} = entity.__struct__.__schema__(:association, key)
```

The reason I did not do this is because it broke all of the Serializer tests.  Those tests are using a plain map as their input test data, but in reality I think this would always be a struct backed by an Ecto Schema.  If we want to make that change, we will need to include Ecto as a test dependency, and that was starting to make this PR too large.

PR #298, opened 2023-06-28
Format required checks for relationships

matt-glover:
This PR updates the JSON:API format verification plug to include checks for the relationships key when POST-ing to create a new resource.

This is intended to help address: https://github.com/beam-community/jsonapi/issues/294

Additionally this PR:
1. Adjusts the list behavior in `ErrorView.send_error/2` because I don't think it extracted the status code previously
2. Added support for serializing a list of formatting errors to handle multiple malformed relationship entries

### DIFF
--- a/lib/jsonapi/plugs/format_required.ex
+++ b/lib/jsonapi/plugs/format_required.ex
@@ -12,6 +12,57 @@ defmodule JSONAPI.FormatRequired do
 
   def call(%{method: method} = conn, _opts) when method in ~w[DELETE GET HEAD], do: conn
 
+  def call(
+        %{method: "POST", params: %{"data" => %{"type" => _, "relationships" => relationships}}} =
+          conn,
+        _
+      )
+      when not is_map(relationships) do
+    send_error(conn, relationships_create_missing_object())
+  end
+
+  def call(
+        %{
+          method: "POST",
+          params: %{"data" => %{"type" => _, "relationships" => relationships}}
+        } = conn,
+        _
+      )
+      when is_map(relationships) do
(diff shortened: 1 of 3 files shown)


matt-glover:
Was commenting about some caveats and rereading docs and I _think_ I discovered something driving a change here. Closing for now.

lib/jsonapi/plugs/format_required.ex
```diff
@@ -12,6 +12,57 @@ defmodule JSONAPI.FormatRequired do
 
   def call(%{method: method} = conn, _opts) when method in ~w[DELETE GET HEAD], do: conn
 
+  def call(
+        %{method: "POST", params: %{"data" => %{"type" => _, "relationships" => relationships}}} =
```
matt-glover, line 16: I initially limited this to `POST` because the docs seem pretty clear about the spec/format for POST requests.

From <https://jsonapi.org/format/#crud-creating>:
> If a relationship is provided in the relationships member of the [resource object](https://jsonapi.org/format/#document-resource-objects), its value MUST be a relationship object with a data member. The value of this key represents the [linkage](https://jsonapi.org/format/#document-resource-object-linkage) the new resource is to have.

Whereas PATCH was a lot more indirect. It talks about "relationship object with a data member" which ultimately, in another section of the docs, translates to a resource linkage. So I'll have to update this to support both request types.

At first I was worried about being overly strict because I thought the PATCH left it open to pass stuff other than resource links. I didn't realize "relationship with a data member" was the same as the resource linkage stuff directly linked in the create section.

Future/updated version should address this.

PR #299, opened 2023-06-28
Format required checks for relationships

matt-glover:
This PR updates the JSON:API format verification plug to include checks for the relationships key when creating and updating resources.

This is intended to address: https://github.com/beam-community/jsonapi/issues/294

Additionally this PR:
1. Adjusts the list behavior in `ErrorView.send_error/2` because I don't think it extracted the status code previously
2. Added support for serializing a list of formatting errors to handle multiple malformed relationship entries

### DIFF
--- a/lib/jsonapi/error_view.ex
+++ b/lib/jsonapi/error_view.ex
@@ -7,7 +7,9 @@ defmodule JSONAPI.ErrorView do
   @crud_message "Check out http://jsonapi.org/format/#crud for more info."
   @relationship_resource_linkage_message "Check out https://jsonapi.org/format/#document-resource-object-linkage for more info."
 
-  @spec build_error(binary(), pos_integer(), binary() | nil, binary() | nil) :: map()
+  @type error_attrs :: map()
+
+  @spec build_error(binary(), pos_integer(), binary() | nil, binary() | nil) :: error_attrs()
   def build_error(title, status, detail, pointer \\ nil, meta \\ nil) do
     error = %{
       detail: detail,
@@ -105,8 +107,8 @@ defmodule JSONAPI.ErrorView do
     |> serialize_error
   end
 
-  @spec missing_relationship_data_param_error(binary()) :: map()
-  def missing_relationship_data_param_error(relationship_name) do
+  @spec missing_relationship_data_param_error_attrs(binary()) :: error_attrs()
+  def missing_relationship_data_param_error_attrs(relationship_name) do
     "Missing data member in relationship"
     |> build_error(
       400,
@@ -115,8 +117,8 @@ defmodule JSONAPI.ErrorView do
     )
   end
 
-  @spec missing_relationship_data_id_param_error(binary()) :: map()
-  def missing_relationship_data_id_param_error(relationship_name) do
+  @spec missing_relationship_data_id_param_error_attrs(binary()) :: error_attrs()
+  def missing_relationship_data_id_param_error_attrs(relationship_name) do
     "Missing id in relationship data parameter"
     |> build_error(
       400,
@@ -125,8 +127,8 @@ defmodule JSONAPI.ErrorView do
     )
   end
 
-  @spec missing_relationship_data_type_param_error(binary()) :: map()
-  def missing_relationship_data_type_param_error(relationship_name) do
+  @spec missing_relationship_data_type_param_error_attrs(binary()) :: error_attrs()
+  def missing_relationship_data_type_param_error_attrs(relationship_name) do
     "Missing type in relationship data parameter"
     |> build_error(
       400,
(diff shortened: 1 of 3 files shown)


lib/jsonapi/error_view.ex
```diff
@@ -93,12 +94,57 @@ defmodule JSONAPI.ErrorView do
...
+    |> build_error(
+      400,
+      "Check out https://jsonapi.org/format/#document-resource-object-relationships for more info.",
+      "/data/relationships"
+    )
+    |> serialize_error
+  end
+
+  @spec missing_relationship_data_param_error(binary()) :: map()
+  def missing_relationship_data_param_error(relationship_name) do
```
matt-glover, line 109: I gave the next few functions a `_error` suffix to distinguish them from the other calls that all chain into `serialize_error`. In these cases I want an unserialized result so I can later collect them into a single list of errors to render down in `serializer_errors/1`.

Didn't feel great about the naming. Feels kinda easy to confuse with the functions that do return a full `%{errors: [_]}` response. But this still felt better to me than shuffling around the other code or moving the errors out just for this use case.

lib/jsonapi/error_view.ex
```diff
@@ -93,12 +94,57 @@ defmodule JSONAPI.ErrorView do
...
+      "/data/relationships/#{relationship_name}/data/type"
+    )
+  end
+
   @spec send_error(Plug.Conn.t(), term()) :: term()
   def send_error(conn, %{errors: [%{status: status}]} = error),
     do: send_error(conn, status, error)
 
   def send_error(conn, %{errors: errors} = error) when is_list(errors) do
-    status = Enum.max_by(errors, &Map.get(&1, :status))
```
matt-glover, line 101: Best I can tell this was not called on the latest stable version prior to this PR. I think this is a safe change as long as it works in the new workflows.

lib/jsonapi/plugs/format_required.ex
```diff
@@ -12,6 +12,57 @@ defmodule JSONAPI.FormatRequired do
...
+          error = missing_relationship_data_type_param_error(relationship_name)
+          [error | acc]
+
+        {relationship_name, %{"data" => %{}}}, acc ->
+          id_error = missing_relationship_data_id_param_error(relationship_name)
+          type_error = missing_relationship_data_type_param_error(relationship_name)
+          [id_error | [type_error | acc]]
+
+        {_relationship_name, %{"data" => _}}, acc ->
+          # Allow things other than resource identifier objects per https://jsonapi.org/format/#document-resource-object-linkage
```
matt-glover, line 51: Mostly I don't have a strong personal need to try and validate the list / null cases super thoroughly right now so I left them as a passthrough for now but presumably those could be validated eventually as well.

test/jsonapi/plugs/format_required_test.exs
```diff
@@ -136,6 +136,291 @@ defmodule JSONAPI.FormatRequiredTest do
     refute conn.halted
   end
 
+  test "halts and returns an error for a specified relationships param missing an object value" do
```
matt-glover, line 139: I copy/pasted a lot of the tests below and just swapped in POST vs PATCH. Looks like I missed one in my haste.

Happy to do that here but I'll wait for some review feedback in case reviewers prefer `describe` blocks or tests that check both POST and PATCH in one case or something.

If the format used elsewhere below is fine I can add a PATCH case here also.

lib/jsonapi/plugs/format_required.ex
mattpolzin: I don't think you need to add all related validations for this PR to be a win, but could you mention the unvalidated cases by name in this comment? I know what it means, but it kind of reads as "the spec allows all other structures here" instead of "there are a couple of other valid structures here".

test/jsonapi/plugs/format_required_test.exs
mattpolzin: Personally I would be inclined to use property based testing to cover POST and PATCH in one test case, but that's not going to block approval of this PR -- you can take your pick.

mattpolzin (commented):
This is a great improvement to error handling. Excited to merge once we've settled the last couple of things!

lib/jsonapi/error_view.ex
matt-glover: Pushed up the suffix and new type alias.

lib/jsonapi/plugs/format_required.ex
matt-glover: Copy/pasted comments with the 3 (of 4) other allowed object shapes from the docs

test/jsonapi/plugs/format_required_test.exs
matt-glover: Added the PATCH case.

matt-glover:
Best I can tell some sort of build cache error caused linting to fail. Seems to lint fine for me locally.

mattpolzin (approved):
Thanks! Big improvement to error handling in the given malformed relationship cases.

PR #310, opened 2024-02-12
Add options to transform field keys non-recursively (#132)

protestContest:
Following up on [this discussion](https://github.com/beam-community/jsonapi/issues/132#issuecomment-1928563844).

This adds two new values for the config option `transform_fields`:

- `camelize_shallow`: Like `camelize`, but only affects the relationship and attribute keys, not their values.
- `dasherize_shallow`: Like `dasherize`, but only affects the relationship and attribute keys, not their values.

Some users may have views with fields that are maps, and don't want the map's keys dasherized or camelized. The new config values allow these users to use the `transform_fields` option to transform only the field names, but not the values.

Feedback is requested. I don't really like the config value names—preferably we can think of something better. If this seems like the right direction, I'll continue to update the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,7 +141,7 @@ Transforming fields requires two steps:
 
    ```elixir
    config :jsonapi,
-     field_transformation: :camelize # or dasherize
+     field_transformation: :camelize # or :dasherize, :camelize_shallow, or :dasherize_shallow
    ```
 
 2. Underscoring _incoming_ params (both query and body) requires you add the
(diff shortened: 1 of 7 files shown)


mattpolzin:
This seems like a good direction to go in. I'll do a more in-depth look at your implementation of `expand_fields_once` after you've wrapped up your first draft.

protestContest:
Ok, I've updated the config values to `:camelize_shallow` and `:dasherize_shallow` and updated the docs around this config option. I also included a couple minor unrelated fixes to appease the linter.

mattpolzin (changes requested):
This feels a bit pedantic, but I noticed that `expand_fields/2` handles all sorts of data types whereas `expand_fields_once/2` only handles maps. On one hand, it only needs to in order to handle the job of transforming only the names of attributes of the map within JSON:API `data`. On the other hand, it's a utility function with theoretically broader usage that doesn't do the same thing as `expand_fields/2` but only once as its name suggests.

Any other ideas what to call it? Off the cuff I'd say `expand_map_fields_once/2` but I'm not tied to that name.

Aside from that small detail, this looks good!

protestContest:
Totally. What do you think of `expand_keys`? That sounds more descriptive to me.

mattpolzin:
`expand_root_keys`? I'm good with `expand_keys` too since it no longer sounds like it is as closely related to `expand_fields`, but `expand_keys` would need some code-documentation explaining that it only looks at keys in the root of the map (perhaps that documentation is useful regardless of the name).

mattpolzin (approved):
This looks great. Thank you!

mattpolzin:
I'll cut a new release shortly. https://github.com/beam-community/jsonapi/pull/311

PR #315, opened 2024-02-22
Transform relationship keys with shallow field transformation options (#314)

protestContest:
Closes #314.

This makes the name of `expand_root_keys/2` make less sense, so I'm open to alternative suggestions.

PR #316, opened 2024-02-23
Fix bug where underscore/dasherize misses single characters

protestContest:
This fixes a bug in `JSONAPI.Utils.String.underscore/1` and `JSONAPI.Utils.String.dasherize/1`, where dashes or underscores would be missed when following a single character.

### DIFF
--- a/lib/jsonapi/utils/string.ex
+++ b/lib/jsonapi/utils/string.ex
@@ -33,11 +33,14 @@ defmodule JSONAPI.Utils.String do
       iex> underscore("corgiAge")
       "corgi_age"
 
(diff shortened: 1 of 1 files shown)


protestContest:
I didn't create an issue for this bug, but I can if you'd like it for tracking purposes.

I can also fold this into #315, along with a version bump, if that's easier too.

mattpolzin (approved):
Yay for Regex. Hah. Nicely done.

mattpolzin:
I'll merge this, then let's rebase your other branch and incorporate both changes in the same version and release notes per your suggestion.

PR #333, opened 2024-08-19
fix: [#334] Fix incorrect ordering of serialized lists

protestContest:
A change introduced in `JSONAPI.Serializer` [here](https://github.com/beam-community/jsonapi/compare/1.7.1...v1.8.0#diff-a93b54749b1587f721f6b1b699cfe8e56134c39b70e5b8245a72c71ab550229fR53) (and a couple nearby spots) incorrectly orders a list response.

This fixes the list reversal, or reverts it in some places. List concatenation should be efficient enough in this instance.

Closes https://github.com/beam-community/jsonapi/issues/334

### DIFF
--- a/lib/jsonapi/serializer.ex
+++ b/lib/jsonapi/serializer.ex
@@ -48,10 +48,13 @@ defmodule JSONAPI.Serializer do
   def encode_data(_view, nil, _conn, _query_includes, _options), do: {[], nil}
 
   def encode_data(view, data, conn, query_includes, options) when is_list(data) do
-    Enum.map_reduce(data, [], fn d, acc ->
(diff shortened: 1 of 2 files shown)


mattpolzin:
Makes sense to me. Thanks for the new test that'll catch this regression going forward.

protestContest:
@mattpolzin I pushed some updates to satisfy the CI checks, but it looks like something weird happened with the deps in a couple of them. Can you re-run the checks?

mattpolzin:
Looks all good now. Can you bump the patch version number in this PR to save a bit of time (it'll release right away when we merge this PR that way).

mattpolzin (approved):
Thanks for the fix!

PR #343, opened 2024-10-28
fix: camelize when prefix is camelized already

rodrigues:
This PR fixes a problem when payload relationship is already camelized:

In the POST, when you receive as payload:

```elixir
%{
  type: "example-resource",
  attributes: %{...},
  relationships: %{
    generalManagerProfile: %{data: %{id: manager_id, type: "profiles"}}
  }
}
```

params will resolve to:

```elixir
%{"generalmanagerprofileId" => 139981}
```

with this fix, it resolves correctly to `generalManagerProfileId`, and with the `UnderscoreParameters` plug, becomes `general_manager_profile_id`.

### DIFF
--- a/lib/jsonapi/utils/string.ex
+++ b/lib/jsonapi/utils/string.ex
@@ -109,6 +109,12 @@ defmodule JSONAPI.Utils.String do
       iex> camelize("alreadyCamelized")
       "alreadyCamelized"
 
+      iex> camelize("alreadyCamelized-id")
+      "alreadyCamelizedId"
+
+      iex> camelize("alreadyCamelized_id")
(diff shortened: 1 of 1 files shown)


mattpolzin (commented):
This looks like a good idea to me, but I think it could be argued to regress on something that admittedly is not strictly defined behavior: `ThisThing` used to be transformed into something starting with a lower case `t` but won't anymore, right?

rodrigues:
good point @mattpolzin, thanks, I've pushed a change to address this scenario too:

```elixir
iex> camelize("PascalLambda_id")
      "pascalLambdaId"
```

mattpolzin (approved):

PR #359, opened 2025-03-13
add backwards compatible polymorphic resource support

nduitz:
Hey there,
I tried my best to add support for polymorphic resources.
The idea is to allow the user to infer resource specifics from the given data.
Currently this only allows `type` and `fields` to be polymorphic. I will probably implement relationships as well.

Additionally I only tested it to work as a relation with an api endpoint of its own - quite frankly there are some things that are not possible: 
The `QueryParser` fetches `view.fields()` at a moment where no data is present, thus inference is not possible.

This is my first time really working with a behaviour, my approach on implementing a backwards compatible solution felt a bit clunky, so any feedback is very welcome.  :)

### DIFF
--- a/lib/jsonapi/view.ex
+++ b/lib/jsonapi/view.ex
@@ -140,10 +140,12 @@ defmodule JSONAPI.View do
   @type options :: keyword()
   @type resource_id :: String.t()
   @type resource_type :: String.t()
+  @type polymorphic_resource :: boolean()
 
   @callback attributes(data(), Conn.t() | nil) :: map()
   @callback id(data()) :: resource_id() | nil
   @callback fields() :: [field()]
+  @callback polymorphic_fields(data()) :: [field()]
   @callback get_field(field(), data(), Conn.t()) :: any()
   @callback hidden(data()) :: [field()]
   @callback links(data(), Conn.t()) :: links()
@@ -156,6 +158,7 @@ defmodule JSONAPI.View do
               {atom(), t() | {t(), :include} | {atom(), t()} | {atom(), t(), :include}}
             ]
   @callback type() :: resource_type()
+  @callback polymorphic_type(data()) :: resource_type()
   @callback url_for(data(), Conn.t() | nil) :: String.t()
   @callback url_for_pagination(data(), Conn.t(), Paginator.params()) :: String.t()
   @callback url_for_rel(term(), String.t(), Conn.t() | nil) :: String.t()
@@ -167,7 +170,8 @@ defmodule JSONAPI.View do
     {type, opts} = Keyword.pop(opts, :type)
     {namespace, opts} = Keyword.pop(opts, :namespace)
     {path, opts} = Keyword.pop(opts, :path)
-    {paginator, _opts} = Keyword.pop(opts, :paginator)
+    {paginator, opts} = Keyword.pop(opts, :paginator)
+    {polymorphic_resource, _opts} = Keyword.pop(opts, :polymorphic_resource, false)
 
     quote do
       alias JSONAPI.{Serializer, View}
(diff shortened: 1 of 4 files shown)


mattpolzin:
Thanks for the PR. I’m on a trip for the next two weeks so unless someone else jumps in there will be a bit of a delay in looking this over. Just wanted to let you know I see the PR!

lib/jsonapi/view.ex
```diff
@@ -205,8 +210,21 @@ defmodule JSONAPI.View do
...
 
-      @impl View
-      def fields, do: raise("Need to implement fields/0")
+      cond do
+        !@polymorphic_resource ->
+          @impl View
+          def fields, do: raise("Need to implement fields/0")
+
+          @impl View
+          def polymorphic_fields(_data), do: nil
```
mattpolzin, line 219: The type of this callback does not include `nil` (same for the other new callback).

mattpolzin: If in practice this function just gets ignored whenever `@polymorphic_resource` is `false`, may as well return `[]` (which fits its type spec) instead of `nil`.

mattpolzin: An alternative design might be explicitly allow `nil` and avoid the need for the `@polymorphic_type` boolean by instead defining the `resource_type/1` function to be `polymorphic_type(data)` if it isn't `nil` and otherwise `type()`. Having the `@polymorphic_type` boolean is probably the better option, though.

mattpolzin (changes requested):
This looks like a pretty reasonable approach to me. I'll do a bit more thinking on it but in the meantime there are a few things I found to comment on.

nduitz:
Hey there :) 
I did some refactoring of my first approach.
WDYT?
Additionally I updated the docs to get a better understanding of what is possible and how to use the functionality.

mattpolzin:
Hey! Sorry, I kept not thinking of this PR when I had time to look at it because it had been removed from my review queue after my previous comments. I've added it back into my queue so I should get this reviewed in the next few (week) days.

mattpolzin:
> > This looks pretty good. I only have one inline comment.
> > I do have one high level question, though. Is it better (or more versatile) or just different to do it the way you've got it in this PR -- separate functions for type, fields, and relationships that take data and determine a result based on it -- versus having a mechanism by which a `View` can choose a more specific other `View` module so that you could implement your polymorphism via say 3 distinct `JSONAPI.View` modules (in the old style already supported by this library) and have one new `JSONAPI.View` module that had a single function that took the data and returned the module of the 3 possibilities that should be used for that data.
> 
> Hm I'd have to go over my approach vs the approach you described. Could you give me a short example on how that would look like? I can't seem to find any example for that in the docs, maybe it is maintainer knowledge? :P

Sorry for the delay. I've been busy enough at work that I haven't and still don't have time to reply in detail. The suggestion was not to follow an existing pattern found elsewhere in this library, just a high level alternative idea.

In broad strokes, you could have a callback like the following:
```elixir
@callback polymorphic_specialization(data()) :: module()
```
As an alternative to several callbacks for polymorphic relationships, attributes, etc.

Then you could define "regular" `View` modules for each specialization (by regular I just mean they don't look polymorphic). You would return the correct specialized view module from `polymorphic_specialization/1` and from there the rest of the library's code would operate the same way it always has but using that specialized view instead of a predetermined one.

---

Regardless, don't worry if you don't have time to think on that approach. Instead, give my [inline comment](https://github.com/beam-community/jsonapi/pull/359#discussion_r2152543149) a reply and we can move forward with this PR because it looks like it does the trick and does not break anything and it's a very nice feature to have.

nduitz:
Ahh now I get it, thanks for the clarifaction. Let me have a look, I like the idea of it!

PR #363, opened 2025-05-07
fix: bug with DataToParams include processing expecting more 'data' keys than the spec calls for

mattpolzin:
The JSON:API specification says that the `"included"` key in a response will contain an array of "resource objects" which is to say JSON objects containing `id`, `type`, etc. Prior to this PR, the `DataToParams` module expected `"included"` to contain an array of documents, ostensibly (rooted with `"data"` keys). This PR fixes that handling. The result can be seen clearly in the fixes to the test inputs.

### DIFF
--- a/lib/jsonapi/utils/data_to_params.ex
+++ b/lib/jsonapi/utils/data_to_params.ex
@@ -94,8 +94,8 @@ defmodule JSONAPI.Utils.DataToParams do
 
   defp process_included(%{"included" => included} = incoming) do
     included
-    |> Enum.reduce(incoming, fn %{"data" => %{"type" => type}} = params, acc ->
-      flattened = process(params)
+    |> Enum.reduce(incoming, fn %{"type" => type} = params, acc ->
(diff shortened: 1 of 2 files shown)


epinault:
thanks for this. I looked at https://jsonapi.org/format/#document-resource-objects and aligns with what you said.

epinault (approved):

mattpolzin:
@epinault yeah, I also see it both ways RE breaking vs. not, but when something is non-compliant with a spec I lean in the direction of bug fix.

In this particular case, especially, because this is code that only gets hit when you are handling non-Ecto data (like when acting as a JSON:API client instead of server) which is uncommon for users of this library and I am guessing that is why this bug wasn't caught sooner. If anyone was reliant on this bug, the API they are using would be broken (like to the tune of no includes ever being understood by the client) for any valid JSON:API client of which there are many well tested options out there for most common languages, so I am 98% sure no one is going to see breakages due to this fix.

PR #381, opened 2025-09-04
feat: allow users to configure to skip missing keys

nduitz:
Hey it's me again :)

Maybe this is just a special use case for us but might help others as well.

We use this package in combination with open api.
Open api allows you to configure keys in responses as `required` meaning they might be missing in your response.

In our case this is especially useful because we need to do partial updates.
We could of course send the whole data structure again only updating the relevant keys but that is not very efficient in terms of data usage :)

### DIFF
--- a/lib/jsonapi/view.ex
+++ b/lib/jsonapi/view.ex
@@ -248,6 +248,7 @@ defmodule JSONAPI.View do
     {path, opts} = Keyword.pop(opts, :path)
     {paginator, opts} = Keyword.pop(opts, :paginator)
     {polymorphic_resource?, _opts} = Keyword.pop(opts, :polymorphic_resource?, false)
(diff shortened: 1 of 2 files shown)


mattpolzin:
I think I understand your use-case, but want to be sure. The client is sending the server a partial update and you'd like the server to respond with only the fields that the client request specified as well? The other possibility that occurred to me is that you are using the `JSONAPI.View` to serialize a request payload rather than a response payload in this scenario.

nduitz:
> I think your requirements could also be met with a View implementing the `hidden/1` function. Not sure if we should add a setting that is a shortcut for that existing way of accomplishing what you're after.

```
def hidden(data) do
  View.fields() -- data.keys
end
```

Yup you are right. thank you! Closed.

PR #410, opened 2026-04-06
feat: Add JSON:API sparse fieldset relationship support

mattpolzin:
Fixes #408.

The JSON:API specification allows for sparse fieldsets to limit both attribute and relationship serialization. Until now, this library only appliled sparse fieldset handling to attributes.

This adds relationship support.

Adding this support is a bit fiddly given this library already treats "fields" as a synonym for "attributes" in more than a couple of places. In order to not rip the carpet up, I chose to make this a mostly non-breaking change. I say mostly because there is still one important catch: If an existing API assumes the non-standard behavior of only filtering _attributes_ by the sparse fieldset query parameter, then clients may request a fieldset not including relationship values that they expect to have serialized and this PR will break those API requests (which must now specify the relationship names in the sparse fieldset query parameter going forward).

Still, I consider this a bug fix given the specification is very clear about how sparse fieldsets work. Hopefully no one sees regressions in their own APIs due to this behavioral change.

### DIFF
--- a/lib/jsonapi/plugs/query_parser.ex
+++ b/lib/jsonapi/plugs/query_parser.ex
@@ -282,9 +282,9 @@ defmodule JSONAPI.QueryParser do
   @spec get_valid_fields_for_type(Config.t(), String.t()) :: list(atom())
   def get_valid_fields_for_type(%Config{view: view}, type) do
     if type == view.type() do
-      view.fields()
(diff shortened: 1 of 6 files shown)


mattpolzin:
I don't particularly like self-merging without review, but I'll take a weeks silence as an indicator that no one paying very close attention has any large objections.

Issue #405, opened 2026-03-12
Parent reference in related links


ssbb:
Hi! I am quite new to `jsonapi` so forgive me if I got something wrong (or asking for something wrong) but looking at the spec examples I see this:

```
{
  "data": [{
      "type": "comments",
      "id": "1",
      "attributes": {
          "text": "HATEOS are the thing!"
      },
      "links": {
          "self": "/comments/1"
      },
      "relationships": {
        "author": {
          "links": {
            "self": "/comments/1/relationships/author",
            "related": "/comments/1/author"
          }
        },
        "articles": {
          "links": {
            "self": "/comments/1/relationships/articles",
            "related": "/comments/1/articles"
          }
        }
      }
  }],
  "links": {
      "self": "/comments"
  }
}
```

Few things I noticed when tried to implement similar relationships:

1. `related` link generator unaware of the parent so this lib would generate just `/articles` instead of `/comments/1/articles`. `url_for` on the child view gets just the collection and the conn. But anyway I think it makes more sense to do it on the parent view level similar to `url_for_rel` but for `related` links.
2. There currently no way to generate links without preloading whole association? Looking at the code it seems to skip whole thing if association not preloaded. With this spec example in mind - wouldn't it make more sense to just omit `data` section then but still generate the links?

mattpolzin:
Sorry for not replying to your issue for a really long time. Partly, I have not used the links features of this library so I am hardly familiar with their inner workings.

In broad strokes, your points both make sense. The first point sounds like a legitimate unhandled case for links generation currently. The second point sounds like one of a category of limitations in this library around preloading. For example, a to-one relationship is skipped if not loaded (though you often have an ID around and could conceivably create the `relationship object` just the same). It sounds like you're right here, the link generation could be separated from the logic of checking for serialization in theory. It may not be trivial to make those code changes because the `View`'s `links` callback takes the `data` as its first argument, but if you are handling `nil` as a case there it may not be too big of a change.

grzuy:
Hi 👋 

Kind of facing similar limitations to what you are reporting.

> 2\. There currently no way to generate links without preloading whole association? Looking at the code it seems to skip whole thing if association not preloaded. With this spec example in mind - wouldn't it make more sense to just omit `data` section then but still generate the links?

I think something like https://github.com/beam-community/jsonapi/pull/427 would fix this?

grzuy:
[...]

PR #427, opened 2026-05-19
feat: include relationship object links despite not data

grzuy:
fixes point number 2 of #405


grzuy:
Realized that maybe better to refactor to provide better control over relationships serialization than just patching like I am doing here...

Reporting in https://github.com/beam-community/jsonapi/issues/435.

Issue #285, opened 2023-01-11
Standardize on underscoring query parameters


TylerPachal:
I was working on a PR to add "dasherized support" for the `sort` and `include` fields, in the same way I added it for the `filter` field in https://github.com/beam-community/jsonapi/pull/282.

While I was doing that, I realized that it already works for `include`, because the [`parse_include` function performs its own underscoring already](https://github.com/beam-community/jsonapi/blob/1fb6380c9b53c64f5625fd3238d87564d6e6655b/lib/jsonapi/plugs/query_parser.ex#L142), and doesn't need any help from the `UnderscoreParameters` plug.

This was something that the `parse_sort` function does not do, which I addressed by changing the `UnderscoreParameters` plug in https://github.com/beam-community/jsonapi/pull/282.

So now we are in a state where certain query parameters always _underscore_ their values, while others do not:

| Query Parameter | Underscored? |
| ----- | ----- |
| `fields` | [Always](https://github.com/beam-community/jsonapi/blob/1fb6380c9b53c64f5625fd3238d87564d6e6655b/lib/jsonapi/plugs/query_parser.ex#L142) |
| `filter` | Opt-in via `replace_query_params` |
| `include` |  [Always](https://github.com/beam-community/jsonapi/blob/1fb6380c9b53c64f5625fd3238d87564d6e6655b/lib/jsonapi/plugs/query_parser.ex#L206) |
| `sort` |  Never |

I think for version `2.0` of this library:
- All of the query parameters should always be underscored.  It should be done by the `QueryParser` module.
- The `UnderscoreParameters` module should go back to just doing the `params`, mostly undoing the work in https://github.com/beam-community/jsonapi/pull/282.

But in the mean time, there should still be a way to underscore the `sort` query parameter.  It looks like our options are:

1. Add the `sort` underscoring in the same fashion as the `filter` underscoring by putting it behind the `replace_query_params` option.
2. Do it all of the time by adding a call to `underscore/1` in the `QueryParser.parse_filter` function.

Let me know what you think is best, and I can prepare the PR.

esbanarango:
🙏

grzuy:
> This was something that the `parse_sort` function does not do

FWIW: It does now after https://github.com/beam-community/jsonapi/pull/431.

grzuy:
> All of the query parameters should always be underscored. It should be done by the QueryParser module.

Agree.
https://github.com/beam-community/jsonapi/pull/434.

PR #431, opened 2026-05-22
fix: fixes parsing of sort fields

grzuy:
Hi 👋 

An attempt to partially fix sort issues mentioned in #285 when sort fields are dasherized or camelCased.
Hope is useful to others.


mattpolzin:
This is a breaking change in some ways, but it does align handling with the more common case (includes and fieldsets are already done this way) and since the next step is atomizing it is probably safe to assume that in nearly every case the desire is not to use strange camel-case atoms. Let's go for it and see if it causes problems for folks since that seems really unlikely.

mattpolzin (approved):

grzuy:
Thank you very much 🙏

PR #434, opened 2026-05-22
fix: fixes `QueryParser` parsing of dasherized/camelized filter keys

grzuy:
Analogous to #431 but for `filter`.

`QueryParser` is already automatically underscoring `fields`, `sort` and `include`, just missing `filter`.

Seems this would be the last fix needed to "close" [Standardize on underscoring query parameters #285](https://github.com/beam-community/jsonapi/issues/285).

---

Underscoring of `filter` keys is technically already supported by `UnderscoreParameters` plug, but it feels that this should work for anyone using `QueryParser` and not `UnderscoreParameters` plug.

I kind of agree with the following statement by @TylerPachal statement in #285 

> I think for version 2.0 of this library:

  > All of the query parameters should always be underscored. It should be done by the QueryParser module.
  >   The UnderscoreParameters module should go back to just doing the params, mostly undoing the work in https://github.com/beam-community/jsonapi/pull/282.

### DIFF
--- a/lib/jsonapi/plugs/query_parser.ex
+++ b/lib/jsonapi/plugs/query_parser.ex
@@ -119,6 +119,7 @@ defmodule JSONAPI.QueryParser do
     opts_filter = Keyword.get(opts, :filter, [])
 
     Enum.reduce(filter, config, fn {key, val}, acc ->
+      key = underscore(key)
       check_filter_allowed!(opts_filter, key, config)
 
       keys = key |> String.split(".") |> Enum.map(&String.to_existing_atom/1)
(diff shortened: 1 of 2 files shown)


mattpolzin (approved):
I'm more on the fence about this one because there was a way to get this behavior at least and it constitutes what could be a breaking change for folks but on the grounds that this brings filter query parameters in line with all other query parameter handling I think it is worth accepting nonetheless.

I'd argue for this being part of a v2 release, but I don't have time to put that release together (which should involve revisiting a handful of other deferred breaking changes) so we can move forward with this as a ~minor~ patch release.

grzuy:
OK.
Thank you @mattpolzin

PR #445, opened 2026-07-01
chore: sync files with beam-community/common-config

thebeambot:


mattpolzin:
this regresses on some version updates that have not been merged into the common config yet.

